Issue #268, opened 2012-03-24
Wrap all reflection-related code in a single class (hubs)


pszmyd:
I was trying to get rid of the reflection over Hub types, injecting my own hub and hub action resolving logic.
When I dug deeper it seems it would involve rewriting almost **every** class in current SignalR hubs implementation, which is not one would expect.

I guess it would be much better to provide a **single extension point**, like `IHubManager`, with default implementation -`DefaultHubManager`, operating only on string names of hubs instead of `Type` objects. This approach has a lot of pros, eg.:
- Provides a single extension point
- Cleans up the code - reflection code is kept in a single place.
- Performance - the results of reflection operations can be easily cached

Such manager class could provide methods like eg.:
- `IHub GetHub(string hubName)`
- `IEnumerable<IHub> GetHubs()`
- `IEnumerable<HubActionInfo> GetHubActions(string hubName)`
- `HubActionInfo GetHubAction(string hubName, string actionName)`

etc., where `HubActionInfo` object keeps necessary data about a given method (name, names and types of parameters), but _not_ the `MethodInfo` itself (for keeping the manager totally disconnected from reflection stuff).

I could commit to do such refactoring.

PR #275, opened 2012-03-27
Dynamic hubs initial implementation.

pszmyd:
As discussed here: https://github.com/SignalR/SignalR/issues/268

There are lots of file changes, but the default behavior stays the same. The main change is the addition of **Descriptor** objects: 
- `HubDescriptor`, which holds information about the hub (name and type) 
- `ActionDescriptor`, which holds information about the action and a delegate for easy invoking
- `ParameterDescriptor`, holding info about a parameter (name and type) 

and appropriate **descriptor provider** interfaces, along with their default implementations:
- `IHubDescriptorProvider`, with default `ReflectedHubDescriptorProvider` class
- `IActionDescriptorProvider`, with default `ReflectedActionDescriptorProvider` class

_These interfaces add important *_extension points*\* - there can exist **multiple implementations** in addition to reflection-based default ones. It gives the possibility to eg. attach actions to hubs at runtime, dynamically create hubs etc. Generally speaking - **hub and action descriptors can come from different sources**. 
The only thing you have to do is to register your custom provider with the dependency resolver.*

**The core class, `DefaultHubManager` (implementing `IHubManager`) resolves available providers and uses their merged outputs to provide it's services.**

All necessary classes (hub dispatcher, JS generator etc.) were refactored to make use of `IHubManager`. 

_Unit tests (`DefaultActionResolverTest`) were also changed to utilize the new classes. All tests are being passed. The current implementation needs some more unit tests, which I'll provide in next few days. I've thoroughly tested changes using all of the included samples and everything works perfectly._

---

I've also added some fixes and performance tweaks:
1. **Peformance** - resolving an action takes ~20x less time in comparison to current version (resolving an action 100k times in a loop took **~120ms, in comparison to ~3000ms** as we have now). I've added a performance test to `DefaultActionResolverTest`. This is mainly because of dictionary-based caching.
2. **Fixes** - there are a few, eg. moving the ASP.NET-specific type resolver registration code from handler to the bootstrapper for performance.

### DIFF
--- a/SignalR/Hubs/HubDispatcher.cs
+++ b/SignalR/Hubs/HubDispatcher.cs
@@ -63,7 +63,7 @@
             try
             {
                 // Invoke the action
-                // pszmyd: Invoker delegate automatically adjusts JSON parameters to correct types.
+                // pszmyd: Invoker delegate automatically deserializes JSON parameters and adjusts them to the given action.
                 object result = actionDescriptor.Invoker(hub, parameters);
                 Type returnType = result != null ? result.GetType() : actionDescriptor.ReturnType;
 
--- a/SignalR/Hubs/Lookup/Descriptors/HubDescriptor.cs
+++ b/SignalR/Hubs/Lookup/Descriptors/HubDescriptor.cs
@@ -10,11 +10,11 @@ namespace SignalR.Hubs.Lookup.Descriptors
         /// <summary>
         /// Hub name.
         /// </summary>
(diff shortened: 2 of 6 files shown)


atlantim:
I have two hubs named MainHub and TempHub. When I want to turn off TempHub using the Dispose method, it is also affected by MainHub and all clients on MainHub are ReConnected.

I can not find the reason for this.

Issue #1744, opened 2013-03-23
IE 10 Web Sockets Security Error when simultaneous connection limit reached


dangreendotnet:
I have found that when using websockets IE 10 throws SCRIPT5022: SecurityError if more that 6 pages with signalr are opened using window.open().

If you click the open new window button on the Default.aspx page the 7th SignalrPage.aspx page will throw the error.
If you just load the SignalrPage.aspx page in new windows, you can open as many windows as you want without error.

I have created a repo project and tested this on windows server 2012 and IE 10 on the same server.

Here is a link to my repo project
https://skydrive.live.com/redir?resid=453C883288932B34!7350

PR #2112, opened 2013-06-04
1744 - Addressed WebSocket security error in IE10

NTaylorMullen:
When IE10 had more than 6 pages open on the same domain with WebSockets it would throw a SecurityError.  We now catch this error appropriately (and any others that might be thrown by a transport starting) and fallback appropriately.

I had to also expand the network mock library to allow for enabling/disabling the library so that I could test the native dom WebSocket element (need to test the original socket, not the modified one).

This was all to address Issue #1744


manuish02:
I am facing the same issue in I.E11 :( :( I am trying to open multiple websocket connections on the same domain name and it causes issue after certain number of pages(this is also not fixed). When I debug it throws me the Security Error Exception.

Issue #2774, opened 2013-12-18
Signal R - Large Messages are not being passed from the client to server


smit010:
We get the "The Uri string is too long." error when passing a large message from the client back to the server.

![uri_error](https://f.cloud.github.com/assets/3926319/1778961/1b95fb04-683e-11e3-8a5b-09211645bf35.jpg)

The function in the .NET client which sends data back to the server:

![senddata](https://f.cloud.github.com/assets/3926319/1779054/9aa82010-683f-11e3-9e5e-a28a23a2fbd6.jpg)

ENVIRONMENT:
Signal R Server on Windows 2012 server/Windows 7.
Signal R .NET Client on Windows 7 machines.

We use async calls, but as far as we understand it, we need async calls for compatibility with windows 8 and windows server 2012 machines.

Also, the issue appeared as a result of upgrading from signal R release 1.1.0 to 2.0.1 and no other changes were made. Async calls were added much before this upgrade.

PR #2829, opened 2014-01-23
Make the max client-to-server WebSocket message size configurable 

halter73:
- The max message size can be set/disabled via the IConfigurationManager
#2774


timforever:
Thanks for making this change, it has helped me out!

I'm using a SignalR clients to send JPEG images to create a sort of video stream. I realize this is probably way out of the scope of the intended use of SignalR, but hey, until I get a chance to create a better solution, its working!

PR #3368, opened 2014-12-25
Fix issue with dynamic assemblies. Calling the Location property on a dy...

achibus:
...namic assembly throws a NotSupportedException.

Suggestion: Check if the assembly is dynamic before using the location property


dstarcev:
Why it's still not merged?

achibus:
What can I do more that this simple change is getting merged? I would need this in our production app!

Lette:
**Bump**

However, I'd suggest changing `"na"` to something like `"dynamic assembly"` instead, for clarity.

sklose:
I can confirm this as a bug that crashed our production service last night after a deployment. We are dynamically creating proxy classes for rest service calls during startup and it seems due to bad timing the first SingalR connection was established somewhere between generation of that proxy and it's first usage. During SingalR initialization the ReflectedHubDescriptorProvider.GetTypesSafe then triggered an exception for the dynamic assembly and the catch clause triggered yet another exception when it's trying to write to the trace. My current workaround for this issue is a custom IAssemblyLocator implementation (see below), but it would be great if this fix could be merged in anyways.

```
    public class AssemblyLocator : IAssemblyLocator
    {
        public IList<Assembly> GetAssemblies()
        {
            return AppDomain.CurrentDomain.GetAssemblies().Where(a => !a.IsDynamic).ToList();
        }
    }
```

bklooste:
This doesnt cover other issues you can get with reflection only Location. eg an Assembly can have Not supported on any field and it should not kill SIgnalR.

saasen:
@sklose: Is it enough to have that implementation anywhere in the solution? Or do you need to provide a customer Resolver to SignalR in order to register it?

datnguyentps:
@saasen You need to register the custom assembly locator in IoC container and SignalR will use it. Reference: https://stackoverflow.com/questions/17063891/signalr-iassemblylocator-override-not-firing.

saasen:
@datnguyentps: Thanks. I found out myself. I forgot to inform you guys about that. I did exactly the same as you just linked! 👍

anurse:
Closing this as it got stale, sorry :). We actually ended up making the fix though!

https://github.com/SignalR/SignalR/blob/98c53dd9aed814f9265ce4061772b026043b07e0/src/Microsoft.AspNet.SignalR.Core/Hubs/Lookup/ReflectedHubDescriptorProvider.cs#L113

PR #3378, opened 2015-01-06
fixed a major performance bottleneck in BufferTextWriter, attempted same for ChunkBuffer

BrannonKing:
`BufferTextWriter`'s most common call was the `Write(char[], int, int)` method. This method was not overridden for efficient handling. It was breaking that down into individual bytes. Overriding that method shows a 20x speed increase in my testing with 40k-long messages. Also, I don't think it was right for that method to not attempt a flush after the last character on a given write. It was waiting until the next write call came in to flush the previous data.

I also changed `ChunkBuffer` to use List instead of `StringBuilder`. That seems to have better performance on the indexer. However, `ChunkBuffer` and its unit tests seem wrong to me in general. We shouldn't have to scan the entire buffer for newline characters. The sentinel should be at the end of one of the incoming chunks. We should only have to check the last character in each of the `ChunkBuffer.Add` methods. If we could do that it would eliminate most of the time on the read call, according to my profiling. 

image

image

image

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEvents/ChunkBuffer.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEvents/ChunkBuffer.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.md in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Text;
 
 namespace Microsoft.AspNet.SignalR.Client.Transports.ServerSentEvents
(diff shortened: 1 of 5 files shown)


BrannonKing:
I realized that we don't technically need to flush the last packet inside the buffer's write methods because we always manually call flush when we are done writing to the buffer.

halter73:
[In our SignalR 3.00. repo](https://github.com/aspnet/SignalR-Server/blob/3.0.0-beta1/src/Microsoft.AspNet.SignalR.Server/Infrastructure/Connection.cs#L230), we've removed `BufferTextWriter` and `BinaryTextWriter` and replaced it with [MemoryPoolTextWriter](https://github.com/aspnet/SignalR-Server/blob/3.0.0-beta1/src/Microsoft.AspNet.SignalR.Server/Infrastructure/MemoryPoolTextWriter.cs).

I think the same performance issue still exists in the new code since it doesn't not provide an override for `TextWriter.Write(char[] buffer, int index, int count)` leading to the same degenerate behavior where  `TextWriter.Write(char)` is called repeatedly for each character instead of efficiently copying the array using `Array.Copy`.

It would be great if you could port your fix to the `MemoryPoolTextWriter` in the 3.0.0 repo. I think it is better to make the fix there, then we could backport the memory pool code to 2.x. This way we don't have to maintain two variations of this fix.

Also be sure not to mix tabs and spaces :wink:.

BrannonKing:
@halter73 , I downloaded the 3.0 repository and opened it in VS2015 Preview. I made the change to add the override easily enough, but I cannot get the unit tests to run. Well, I can run them all and 40 of them fail with errors about missing dependencies. If I try to run any individually I get this error: An exception occurred while invoking executor 'executor://kunittestexecutor/v1': Unable to run the test . It looks like there are bugs with the dependency copy in the preview version. Here is the override:

```csharp
        public override void Write(char[] value, int index, int length)
        {
            var sourceIndex = index;
            var sourceLength = length;
            while (sourceIndex < sourceLength)
            {
                if (_textLength == _textEnd)
                {
                    Encode(false);
                }

                var count = sourceLength - sourceIndex;
                if (count > _textLength - _textEnd)
                {
                    count = _textLength - _textEnd;
                }

                Array.Copy(value, sourceIndex, _textArray, _textEnd, count);
                sourceIndex += count;
                _textEnd += count;
            }
        }
```

halter73:
@BrannonKing Maybe you can try installing the KVM and KRE.

https://github.com/aspnet/Home/tree/dev#windows-1

You might also need to add the following to you NuGet package sources: https://www.myget.org/F/aspnetvnext/api/v2

Once you do this, you should be able to run the build script which will also run the tests.

If you want to start a PR over on the https://github.com/aspnet/SignalR-Server repo with what you think works, we can continue this discussion over there.

BrannonKing:
I've added the pull request to 3.0. It appears that the 3.0 client is a long way off. For that reason, I'm going to close this pull request and reopen one here on the 2.2 code. (And i'll fix the tab usage for that request.) Let's do a 2.2.1 with the current pull requests.

Issue #3376, opened 2015-01-02
occasional null ref in ServerSentEventsTransport.<OnStart>


BrannonKing:
I'm occasionally seeing the exception shown below in my application logs. I cannot tell that it has a negative impact nor what is causing it. I have no idea how to reproduce it at will.

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNet.SignalR.Client.Transports.ServerSentEventsTransport.<OnStart>b__0()
   at System.Action.Invoke()
   at Microsoft.AspNet.SignalR.Client.Infrastructure.TransportInitializationHandler.<>c__DisplayClassa.<CompleteFail>b__9()
   at Microsoft.AspNet.SignalR.Client.Infrastructure.TransportInitializationHandler.<>c__DisplayClassd.<Dispatch>b__c(Object _)
   at System.Threading.QueueUserWorkItemCallback.WaitCallback_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()
```

moozzyk:
What version are you using? Can you post logs?

BrannonKing:
I'm using version 2.1.2. I'll see if I can work up a better log file. I assume that you only need client-side logs?

moozzyk:
Yes, only client logs.

BrannonKing:
Here's a log file:
http://kingtown.net/Uploads/Mobius_2015-01-02--13-03-30.log

BrannonKing:
I added a null check to my branch:
https://github.com/BrannonKing/SignalR/commit/646e7d6a8b65dae33385b2be3cec527eba63bc0a

I'm guessing a little bit at the location.

moozzyk:
[...]

PR #3385, opened 2015-01-15
20x speedup for BufferTextWriter

BrannonKing:
I posted this pull request once before, but I've now updated it for 2.2 and fixed the spacing. The JSON serializer uses the TextWriter's `Write(char[], int, int)` overload.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
@@ -53,7 +53,9 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
             // if the transport failed to start we want to stop it silently.
             _stop = true;
 
-            _request.Abort();
+            var request = _request;
+            if (request != null)
+                request.Abort();
         }
 
         private void Reconnect(IConnection connection, string data, CancellationToken disconnectToken)
@@ -95,10 +97,10 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
 
             var getTask = HttpClient.Get(url, req =>
             {
+                req.Accept = "text/event-stream";
                 _request = req;
-                _request.Accept = "text/event-stream";
 
-                connection.PrepareRequest(_request);
+                connection.PrepareRequest(req);
 
             }, isLongRunning: true);
 
@@ -107,7 +109,9 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
                 _stop = true;
 
                 // This will no-op if the request is already finished.
-                ((IRequest)state).Abort();
+                var request = state as IRequest;
+                if (request != null)
+                    request.Abort();
             }, _request);
             
             getTask.ContinueWith(task =>
@@ -200,10 +204,9 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
 
         public override void LostConnection(IConnection connection)
         {
-            if (_request != null)
-            {
-                _request.Abort();
-            }
+            var request = _request;
+            if (request != null)
+                request.Abort();
         }
     }
 }
--- a/src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
@@ -55,7 +55,9 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
 
             var request = _request;
             if (request != null)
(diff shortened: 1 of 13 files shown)


halter73:
Previous PR: https://github.com/SignalR/SignalR/pull/3378

src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
```diff
@@ -200,10 +204,9 @@ internal void OpenConnection(IConnection connection, string data, CancellationTo
 
         public override void LostConnection(IConnection connection)
         {
-            if (_request != null)
-            {
-                _request.Abort();
-            }
+            var request = _request;
+            if (request != null)
```
davidfowl: Keep the style

src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
```diff
@@ -107,7 +109,9 @@ internal void OpenConnection(IConnection connection, string data, CancellationTo
                 _stop = true;
 
                 // This will no-op if the request is already finished.
-                ((IRequest)state).Abort();
+                var request = state as IRequest;
+                if (request != null)
```
davidfowl: When would it be null?

src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
```diff
@@ -53,7 +53,9 @@ protected override void OnStartFailed()
             // if the transport failed to start we want to stop it silently.
             _stop = true;
 
-            _request.Abort();
+            var request = _request;
+            if (request != null)
```
davidfowl: When would it be null?

src/Microsoft.AspNet.SignalR.Core/Infrastructure/PerformanceCounterManager.cs
```diff
@@ -438,6 +438,14 @@ public IPerformanceCounter LoadCounter(string categoryName, string counterName,
             // and when they are thrown. 
             try
             {
+                if (!PerformanceCounterCategory.Exists(categoryName))
```
davidfowl: We removed the Exists check for a reason. Please remove it.

davidfowl:
This definitely needs unit tests

src/Microsoft.AspNet.SignalR.Core/Infrastructure/PerformanceCounterManager.cs
BrannonKing: I added the check as I have some developers on my team who disable the "just my code" option from time to time. This exception was causing them some grief. I don't believe that having the counters not installed is an unexpected scenario. Help me understand why the check is bad?

src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
BrannonKing: I don't know when it would be null. I was attempting to fix this issue: https://github.com/SignalR/SignalR/issues/3376 . I was also trying to make the file consistent; I believe that there are null checks for it elsewhere in the file.

src/Microsoft.AspNet.SignalR.Client/Transports/ServerSentEventsTransport.cs
BrannonKing: Issue https://github.com/SignalR/SignalR/issues/3376 shows that it can be null, but I don't know when.

moozzyk: This should be a separate PR. The NRE happens if the disconnectToken is tripped before sending the connect request.

BrannonKing:
I've added a unit test for verification of the method behavior, but not verified that the method is used when appropriate. I couldn't figure out a way to accomplish that latter goal. I also added braces to fix the styling.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/PerformanceCounterManager.cs
rjvdboon: The Exists check can be very expensive, look at the Remarks section of the PerformanceCounterCategory.Exists member on msdn. It's much faster to try to open and fail than to check in advance.

moozzyk:
I believe this was ported to 2.2.1

PR #3392, opened 2015-01-20
.Net 3.5 client support

jbruening:
With AsyncBridge and TaskParallelLibrary, get a working 3.5 client.
Untested.  Should theoretically work though.  Project was basically a direct copy from the 4.5 project, but then just directly reference as if working off original Client project (.net 4) ie, not using websockets, and change back to the .net4 http client stuff

Why 3.5? Unity3d support, basically.  There might be an issue with the Json library's version though, as I was having issues with anything above 5.0.8. AsyncBridge and deps do work in unity though (tested with a different project)


jbruening:
I just ran a test, and receiving appears to be working just fine in unity.  The unity project needs to be changed from .net 2.0 subset to .net 2.0. I was getting messages correctly from the server. The auto transport had resolved to server sent events, after doing some reflecting.

davidfowl:
We're not going to take this because we're not supporting .NET 3.5. It would be great if you could create another repository with this code so that the community can maintain it.

PR #3398, opened 2015-01-28
ReflectedHubDescriptorProvider now also finds internal (and private) hubs

Erwinvandervalk:
I'm working on a project that builds a webapplication out of multiple (discrete) components. We'll ILMerge all internal dependencies into these components. 

SignalR doesn't lend itself well to ILMerging, because the hubs needed to be public. In my mind, that's not needed. If you create a hub, you're doing so to be called from javascript. Regardless if the hub itself is public or private. 

This change allows the hubs to be private as well. 

Note, I do have a workaround: 
http://www.erwinvandervalk.net/2015/01/making-signalr-hubs-internal.html


damianh:
:+1: 

In the longer run, a more extensible approach like WebAPIs `IAssembliesResolver` and `IHttpControllerTypeResolver` would be desirable.

PR #3410, opened 2015-02-13
VB Client

BrennanConroy:


samples/Microsoft.AspNet.SignalR.ClientVB.Samples/Client.vb
```diff
@@ -0,0 +1,49 @@
...
+        Try
+            Dim HubConnection = New HubConnection(url)
+            HubConnection.TraceWriter = _traceWriter
+            HubConnection.TraceLevel = TraceLevels.Events
+
+            Dim hubProxy = HubConnection.CreateHubProxy("chat")
+            'Register client function and behavior
+            hubProxy.On(Of String)("send", Sub(data) HubConnection.TraceWriter.WriteLine(data))
+
+            HubConnection.Start().Wait()
```
moozzyk: Use async/await instead of `.Wait()` - while in this case it does not matter much because it is a Console.App this is a sample and people copy the code and then deadlock themselves.

samples/Microsoft.AspNet.SignalR.ClientVB.Samples/Client.vb
```diff
@@ -0,0 +1,49 @@
...
+    End Sub
+    Public Sub Run(url As String)
+        Try
+            Dim HubConnection = New HubConnection(url)
+            HubConnection.TraceWriter = _traceWriter
+            HubConnection.TraceLevel = TraceLevels.Events
+
+            Dim hubProxy = HubConnection.CreateHubProxy("chat")
+            'Register client function and behavior
+            hubProxy.On(Of String)("send", Sub(data) HubConnection.TraceWriter.WriteLine(data))
```
moozzyk: Writing to log seems weird.

samples/Microsoft.AspNet.SignalR.ClientVB.Samples/Client.vb
```diff
@@ -0,0 +1,49 @@
...
+
+            'User input loop
+            Dim send As String = ""
+            While send <> "/q"
+                hubProxy.Invoke("Send", send).Wait()
+                send = Console.ReadLine()
+            End While
+
+            '-----Examples-----
+            Dim joinGroupResponse = hubProxy.Invoke(Of String)("JoinGroup", HubConnection.ConnectionId, "CommonClientGroup").Result
```
moozzyk: These could be options in the main loop where /j would be join group /l leave group etc. Not hard to do but would feel more real life...

moozzyk:
@BrennanConroy: don't be sad. Here is a gift for you: 🎈

PR #3432, opened 2015-03-19
WeAreSorryHubBase - bringing `HubBase` back to fix binary compatibility

moozzyk:
#3489


moozzyk:
CC: @halter73

moozzyk:
Ping @halter73 @davidfowl

src/Microsoft.AspNet.SignalR.Core/Hub.cs
```diff
@@ -8,68 +8,18 @@ namespace Microsoft.AspNet.SignalR
...
-        protected virtual void Dispose(bool disposing)
-        {
-        }
-
-        public void Dispose()
-        {
-            Dispose(true);
+            get
+            {
+                return HubClients;
```
halter73: Glad we're not reintroducing that stackoverflow bug :grinning:

halter73:
:ship:

src/Microsoft.AspNet.SignalR.Core/Hub.cs
moozzyk: Yeah. One of the goals was to keep the fix :)

Issue #1655, opened 2013-03-08
.Net client throw ArgumentNullException from JToken.FromObject when Invoke hub method with paramater value null


Xiaohongt:
e.g. 
```csharp
//use the asp.net sample url
var hubConnection = new HubConnection("http://localhost:40476/");
var myHub = hubConnection.CreateHubProxy("shapeShare");
hubConnection.Start().Wait();
myHub.Invoke("changeUserName", "test", null).Wait();
```

Results:
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
Parameter name: o
   at Newtonsoft.Json.Linq.JToken.FromObjectInternal(Object o, JsonSerializer jsonSerializer)
   at Newtonsoft.Json.Linq.JToken.FromObject(Object o, JsonSerializer jsonSerializer)
   at Microsoft.AspNet.SignalR.Client.Hubs.HubProxy.Invoke[T](String method, Object[] args) in e:\dd\SignalR_Dev_1.1_2\S
ignalR\src\Microsoft.AspNet.SignalR.Client\Hubs\HubProxy.cs:line 89
   at Microsoft.AspNet.SignalR.Client.Hubs.HubProxy.Invoke(String method, Object[] args) in e:\dd\SignalR_Dev_1.1_2\Sign
alR\src\Microsoft.AspNet.SignalR.Client\Hubs\HubProxy.cs:line 70

Inverness:
Why does this error still exist? It's a simple fix by replacing JToken.FromObject(), which does a null check, with JValue.CreateNull() if the incoming argument is null.

moozzyk:
Fixed in 389427e88eaff56bd9e40bcf7521af4d1c603f07

PR #3449, opened 2015-04-08
Fixed an issue (#1655) with null values failing to serialize.

goss2326:
Fixed an issue (#1655) with null values failing to serialize due to an ArgumentNullException..

JToken.FromObject() throws an argument null exception if the 'o' parameter is null.  JValue however, is a derived type of JToken that has multiple constructors that allow nulls for serialization (including Nullable<T> items).

This is useful as service calls can legitimately allow null values.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs
@@ -1,13 +1,11 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.md in the project root for license information.
 
+using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


goss2326:
@Inverness Thank you for your comment.  After review of Newtonsoft.Json I agreed and firmed that the issue you mentioned would occur (previously I was not using complex objects or collections).  I have made the modification you suggested and retested using this new solution.

imuller:
When will this fix be available in an new release?

moozzyk:
@goss2326 - I rebased on dev, squashed and added a unit test (in a follow-up commit). 389427e88eaff56bd9e40bcf7521af4d1c603f07

@Inverness - good catch noticing the issue with the original fix!

PR #3461, opened 2015-04-17
Added support for nested path with Cors

emadb:
I was facing a case where the Hub was in a nested folder and on a different server. I tried different ways to make it works but I wasn't able to configure the connection Url in the right way.
It seems to me that this case is not supported. The client code that I use is:

``` javascript
$(function () {
  var connection = $.hubConnection('http://remoteWebServer.com/mySubPath/signalr', {useDefaultPath: false});
  connection.start().done(function () {
    console.log('connected');
  });
});
```

Whatever I try the connection fails to negotiate (because mySubPath was removed) or it tries to connect to the same server the page was served (localhost in my case).

So I modified the way in which the baseUrl is created to take care of mySubPath.
Hope that this could be helpful.


aspnet-hello:

Issue #3457, opened 2015-04-15
Ability to specify an assembly for "signalr.exe ghp"


alex-simonov:
In my solution I have two web projects and each contains a hub with name "PortfolioHub". 
I have added `AfterBuild` target in one of my web projects to generate hub proxies:

```
<Target Name="AfterBuild">
  <Exec Command="$(SolutionDir)packages\Microsoft.AspNet.SignalR.Utils.2.2.0\tools\signalr.exe ghp /o:$(MSBuildThisFileDirectory)Scripts\signalR.hubs.js" WorkingDirectory="$(OutDir)" />
</Target>
```

Everything is fine when I build solution in Visual Studio and `signalR.hubs.js` is successfully generated.
But I see an error when I run the same build in TFS:

```
EXEC: System.InvalidOperationException: Two Hubs must not share the same name. 'WebApplication1.Hubs.PortfolioHub, WebApplication1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' and 'WebApplication2.Hubs.PortfolioHub, WebApplication2, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' both share the name 'PortfolioHub'.
```

This is because TFS by default use the same `$(OutDir)` directory for each project in solution and assemblies `WebApplication1` and `WebApplication2` are placed together in it.

It would be great to have an ability to specify an assembly which I want to load for generating hub proxies. For example:

```
signalr.exe ghp /assembly:WebApplication1.dll
```

Thank you!

aspnet-hello:
This issue has been closed as part of issue clean-up as described in https://blogs.msdn.microsoft.com/webdev/2018/09/17/the-future-of-asp-net-signalr/. If you're still encountering this problem, please feel free to re-open and comment to let us know! We're still interested in hearing from you, the backlog just got a little big and we had to do a bulk clean up to get back on top of things. Thanks for your continued feedback!

PR #3469, opened 2015-04-22
Ability to specify an assembly for "signalr.exe ghp"

alex-simonov:
- Argument `/f` was added to `GenarateHubProxyCommand` (`f` means `filter`) - filters assemblies which must be loaded when generating proxies, accepts strings like `Assembly.dll`, `Namepsace.*.dll`, `*.dll` and so on (extension can be ommited, `Namespace.*` is the same as `Namespace.*.dll`);
- Assembly loading algorithm was changed to respect `/f` argument - after filtered assemblies have been loaded all referenced assemblies will be loaded recursively also.
#3457


aspnet-hello:

PR #3480, opened 2015-04-30
ForYouTheStoreIsClosed

### LINKED ISSUE #3393: Cannot install portable NET client package
Hi everyone, 

I was trying to update SignalR client package from 2.1.2 to 2.2, but I got this:
`Install-Package : Could not install package 'Microsoft.AspNet.SignalR.Client 2.2.0'. You are trying to install this package into a project that targets 'portable-net45+sl50+win+wpa81+wp80', but the package does not contain any assembly references or 
content files that are compatible with that framework. For more information, contact the package author.`

I assume, PCL compatibility is broken.

moozzyk:
When a project targeted both net45 and wpa81 or win8 we did not have a subset that could satisfy this. The store websocket transport can only be used if an app/lib targets wpa81 and/or win8 since it is using WinRT APIs. If the app/lib targets any other platform in addition to wpa81/win8 the store websocket cannot be used and therefore we must not install it. Adding wpa81/win8 platforms to the folder containing just the portable version of Microsoft.AspNet.SignalR.Client.dll will do the trick - this folder will be selected if the application targets any other platform in addition to wpa81/win8

Fixes #3393


moozzyk:
Ping @halter73 @davidfowl

halter73:
I'm not a PCL expert. Am I correct to assume that adding the "+win81+wpa81" targets won't remove compatibility for any currently supported platforms?

moozzyk:
I remember talking about this change to @anurse and my understanding was that adding target framework monikers will only broaden applicable platforms.

anurse:
Correct, as long as we are sure the DLL itself will work on all those platforms. However, it would be best to stick with a real Portable Profile. Does this name actually match a numbered Portable Profile?

halter73:
Alrighty then :shipit:

moozzyk:
@anurse - I am not exactly sure how to find the numbered Portable Profile but assuming it exists I am not sure if I would want to use in place of the string we currently have. Looking at the string immediately tells me what this PCL will run - as opposed to the number.

anurse:
Oh, I totally agree we shouldn't use the number. It's more about seeing if there is a profile that meets our needs and using a string that ends up parsing down to that number :). Let me take a look at the profile mappings and see if it matches up.

build/Build.proj
```diff
@@ -81,7 +81,7 @@
     <!-- Only build the following clients if the SDK is installed -->
     <Projects Include="$(ProjectRoot)\src\Microsoft.AspNet.SignalR.Client.Portable\Microsoft.AspNet.SignalR.Client.Portable.csproj">
       <Build Condition="!$(WP8SDKInstalled)">false</Build>
-      <Platform>portable-net45+sl5+netcore45+wp8+wp81</Platform>
+      <Platform>portable-net45+sl5+netcore45+wp8+wp81+win81+wpa81</Platform>
```
anurse: Actually, something smells a little fishy here. `win81` is supposed to map to `netcore451` so it should already be compatible with the old provide (via `netcore45`).

When I create a portable project targeting all these frameworks and pack it I get: `portable-net45+sl50+win+wpa81+wp80` which maps to `.NETPortable,Version=v4.0,Profile=Profile344`.

Can you try that name and see if it fixes the bug? If it does, that should resolve the issue and ensure we are continuing to use "real" portable profiles.

moozzyk:
@anurse - thanks for checking the profile for me. I will try the one you proposed and see if it works.

anurse:
:+1: for the record all I did was create a Portable Class Library, check the boxes and look at the CSProj ;P.

Issue #3426, opened 2015-03-10
SignalR UnobservedTaskException Causing Possible memory leaks


DawtCom:
We are experiencing a memory leak that seems to be leaking in unmanaged heaps to the tune of close to 1gb day in a heavily used application implementing SignalR.  It seems to correlate to The exception I note below.  In windbg we also notice lots of CONTEXT_CONTAINERS left in memory from IISCORE
```
                                         Type name	     Count	Size
                         iiscore!CONTEXT_CONTAINER	    569961	Unknown
          webengine4!MGD_CONNECTION_STORED_CONTEXT	    185320	Unknown
                             iiscore!W3_CONNECTION	    184710	Unknown
               filter!W3_FILTER_CONNECTION_CONTEXT	    151008	Unknown
                          filter!W3_FILTER_CONTEXT	    150553	Unknown
                           iiscore!W3_MAIN_CONTEXT	     20620	Unknown
                          iiscore!W3_CLONE_CONTEXT	     18230	Unknown
                           iiscore!W3_CONTEXT_BASE	     15307	Unknown
                         webengine4!W3_MGD_HANDLER	     11239	Unknown
                           iiscore!REQUEST_CONTEXT	     10834	Unknown
                          iiscore!RESPONSE_CONTEXT	     10833	Unknown

```

I have several dump files also if we need to send them to you for analysis.  With the gflags +ust flags turned on.

We are using the SignalR 2.2 which should be the latest stable version.

Some Stack Traces:

```

10-Mar-2015 07:21:46 - BlueGrass2.Core.UnobservedTaskExceptionLogger - An unobserved task exception has occurred. 
 System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. ---> System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Web.WebSockets.AspNetWebSocket'.
   at System.Web.WebSockets.AspNetWebSocket.get_State()
   at Microsoft.AspNet.SignalR.WebSockets.WebSocketHandler.IsClosedOrClosedSent(WebSocket webSocket)
   at Microsoft.AspNet.SignalR.WebSockets.WebSocketHandler.<<CloseAsync>b__6>d__8.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #3484, opened 2015-05-06
Issue #3426 UnobservedTaskException

deftflux:
This is a fix for the unobserved task exceptions that we are seeing in our logs when an attempt is made to get WebSocket.State after said web socket is disposed and after the close timeout is reached (due to using WhenAny).  We've tested this change in production.


moozzyk:
It was merged to the dev branch with 80e519ad196527d8f85bbb2e1bdebfd0a07393b3

moozzyk:
Thanks for the contribution!

Issue #3481, opened 2015-05-02
Allow creating partitioned topics in ServiceBus scaleout by configuration


pseabury:
The ServiceBusScaleoutConfiguration should include a mechanism to optionally create partitioned Topics for the scaleout.  Empirical data collected over the past few months indicates that ServiceBus is not really a feasible scaleout mechanism due to the number of failed operations - even at low to moderate messaging volumes (order of 100's of ops per hour).

See - https://msdn.microsoft.com/en-us/library/azure/dn520246.aspx

pseabury:
This will require upgrading SignalR to use Azure SDK v2.2 or later.

aspnet-hello:
This issue has been closed as part of issue clean-up as described in https://blogs.msdn.microsoft.com/webdev/2018/09/17/the-future-of-asp-net-signalr/. If you're still encountering this problem, please feel free to re-open and comment to let us know! We're still interested in hearing from you, the backlog just got a little big and we had to do a bulk clean up to get back on top of things. Thanks for your continued feedback!

PR #3494, opened 2015-05-28
Issue #3481 - Ability to create partitioned Topics for ServiceBus backplane

pseabury:
Add ability to configure the usage of partitioned topics for ServiceBus backplane.
 - Updated WindowsAzure.ServiceBus dependency to 2.6 (partitioning available as of 2.2)
 - Added option for partitioning to ServiceBusScaleoutConfiguration
 - Use partitioning option to create topics if they don't already exist.  You canot convert existing ServiceBus entities to use partitioning, they must be created that way.

#3481


davidfowl:
Partitioning breaks the order guarantees we have. Did you test this?

davidfowl:
Closing this out as there was no reply to the question

Issue #3436, opened 2015-03-25
Redis Backplane failure


Barbu666:
Hi,

We're running 3 Azure web role backed by a Redis backplane in production. It's been running very smoothly in the past month but yesterday the backplane went "down" and this error was all over the place in the 3 servers log. Rebooting the servers fixed the problem.

ERR Error running script (call to f_c883e7b82acc1c2a0bb1f2ee805f928afa1c1abb): @user_script:2: user_script:2: attempt to compare boolean with string 

   at Microsoft.AspNet.SignalR.Messaging.ScaleoutStream.Send(Func`2 send, Object state)
   at Microsoft.AspNet.SignalR.Infrastructure.Connection.Send(ConnectionMessage message)
   at Microsoft.AspNet.SignalR.Hubs.HubDispatcher.Outgoing(IHubOutgoingInvokerContext context)
   at Microsoft.AspNet.SignalR.Hubs.HubPipelineModule.<>c__DisplayClass1b.<BuildOutgoing>b__19(IHubOutgoingInvokerContext context)
   at Microsoft.AspNet.SignalR.Hubs.HubPipeline.Send(IHubOutgoingInvokerContext context)
   at Microsoft.AspNet.SignalR.Hubs.SignalProxy.TryInvokeMember(InvokeMemberBinder binder, Object[] args, Object& result)

Any idea what could have caused this?

Thanks

Jehele:
I use Redis 2.8.12, SignalR with Redis 2.2.0
I found that simple restart of Redis windows service leads to the same and it seems that problem is in RedisConnection.RestoreLatestValueForKey. After a change in the lua script, it seems to work (I investigate it using a simple web project).

Does it make sense? Is the script correct?

new version
```
var redisResult = await _connection.GetDatabase(database).ScriptEvaluateAsync(
                   @"
                    local newvalue=-1
                    if redis.call('EXISTS', KEYS[1]) == 1 then 
                        newvalue = tonumber(redis.call('GET', KEYS[1]))
                    end
                    if newvalue < tonumber(ARGV[1]) then
                        return redis.call('SET', KEYS[1], ARGV[1])
                    else
                        return nil
                    end",
         new RedisKey[] { key },
         new RedisValue[] { _latestMessageId });
```
  
Original version:
```
var redisResult = await _connection.GetDatabase(database).ScriptEvaluateAsync(
                   @"local newvalue = redis.call('GET', KEYS[1])
                    if newvalue < ARGV[1] then
[...]

PR #3563, opened 2015-09-28
Update RedisConnection.cs

duszekmestre:
According to issue: https://github.com/SignalR/SignalR/issues/3436

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
+++ b/src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
@@ -88,8 +88,11 @@ namespace Microsoft.AspNet.SignalR.Redis
                 }
 
                 var redisResult = await _connection.GetDatabase(database).ScriptEvaluateAsync(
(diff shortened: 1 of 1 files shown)


longzheng:
+1

src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
```diff
@@ -88,8 +88,11 @@ public Task ScriptEvaluateAsync(int database, string script, string key, byte[]
                 }
 
                 var redisResult = await _connection.GetDatabase(database).ScriptEvaluateAsync(
-                   @"local newvalue = redis.call('GET', KEYS[1])
-                    if newvalue < ARGV[1] then
```
moozzyk: Why not change only the if condition to something like: 

```
if not newvalue or newvalue < ARGV[1] then
```
additional explicit check?

moozzyk:
I rebased this on dev and committed as ece16d8f56f0d586f0aad18f525b8b42045d107d. I also simplified the script in a followup commit. Thanks for your contribution.

PR #3571, opened 2015-10-09
prevent error "unable to get property 'state' of undefined or null"

akiander:
We see the following error a lot coming from client machines: 

**unable to get property 'state' of undefined or null**

When this happens, our signalr connection never recovers, so the clients remain disconnected which causes problems for our application.

This pull request contains a fix for this issue.


akiander:
Adding a fix for another script error we see periodically from clients: 

**unable to get property 'json' of undefined or null reference**

in the following file: 

jquery.signalR.transports.foreverFrame.js

Can you pull in this request to prevent this error going forward?

moozzyk:
@halter73 - can you give me your opinion on this? I am not sure when we would call `changeState` without a valid connection...

aspnet-hello:

PR #3591, opened 2015-11-09
Fixes #3547: uwp compatibility

### LINKED ISSUE #3547: UWP client can't connect
I've prepared a sample project based on a self hosted server, a WPF client that works, same code but using a UWP app client on w10 doesn't work, it can't connect. I've checked and granted also permission on the UWP app for internet e private connection, but it still can't connect to the self hosted server. I'm using versione 2.2.0
This scenario is simple to try, can someone please try it and let me know it they can get it working ?
Thanks

I've found other people having troubles with UWP where they cannot resolve dll, 
my problem looks different, the app loads and try to connect but it fails on connect 

any idea ?

bc3tech:
Fixes #3547 by creating a new UWP lib that uses the same code as the WinRT one. Verified behavior is correct when compiled for .Net Native w/in a UWP project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ jquery.signalR.js
 jquery.signalR.min.js
 xamarin/SignalRPackage/component/lib/mobile/
 tests/Microsoft.AspNet.SignalR.Client.Store.Tests/AppPackages/
+.lock.json
--- a/src/Microsoft.AspNet.SignalR.Client.UWP/Microsoft.AspNet.SignalR.Client.UWP.csproj
+++ b/src/Microsoft.AspNet.SignalR.Client.UWP/Microsoft.AspNet.SignalR.Client.UWP.csproj
@@ -7,7 +7,7 @@
     <ProjectGuid>{AAAE1A6A-3A7E-4EE2-9624-3907FD1CDC1A}</ProjectGuid>
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>Microsoft.AspNet.SignalR.Client.UWP</RootNamespace>
+    <RootNamespace>Microsoft.AspNet.SignalR.Client</RootNamespace>
     <AssemblyName>Microsoft.AspNet.SignalR.Client.UWP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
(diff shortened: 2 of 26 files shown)


mteper:
:+1: Thank you! Do you build and then copy the dll into your project?

I guess to make this PR complete, the new project should probably be integrated into the build?

mteper:
Also, this PR is not likely to get pulled it since it's submitted against the wrong branch. From Contributing.md:
> Our workflow is loosely based on Github Flow. We actively develop in the dev branch. This means that all pull requests by contributors need to be developed and submitted to the dev branch.

bc3tech:
@mteper yes, it's currently working perfectly in a UWP project live in the Windows Store on both Mobile and Big Windows.

I'll re-submit my PR shortly. Sorry for the confusion.

mveroukis:
So I pulled this branch, built the entire solution in Release mode | Any CPU. When I attempt to add a reference to it from my project (Windows 10 UWP) I get the following: 
A Reference to '...\SignalR\src\Microsoft.AspNet.SignalR.Client\bin\Release\Microsoft.AspNet.SignalR.Client.dll' could not be added. The project targets '.NETCore' while the file reference targets '.NETFramework'. This is not a supported scenario.

Anyway how I can get around this? WHen I added this project via NuGet it seemed to work fine in Debug mode.

moozzyk:
@mveroukis - can you try this branch: bug-3547-websockets ? It has some additional fixes to the PR proposed by @bc3tech and also support for the websocket transport.
Pull and then from VS 2015 Command prompt run `msbuild build\build.proj /t:BuildPackages` - this should give you NuGet packages in the artifacts folder. Then try installing the package to your project.

mveroukis:
I don't see any branch by that name. However, I did get past my issue above when I realized that what I wanted was the Microsoft.AspNet.SignalR.Client.Store project. The portable one might work as well though, maybe I should try that.

moozzyk:
@mveroukis - here is the branch https://github.com/SignalR/SignalR/tree/bug-3547-websockets 
I don't believe the Store client will work on UWP. You will be able to install it but it won't work - you will see something like described in #3547.

mveroukis:
Thanks, I will look into that.

bc3tech:
@moozzyk is correct, the Store client won't work which is why this PR (and subsequently #3597) was submitted.

Issue #3594, opened 2015-11-14
TypeLoadException occurring when using Strongly typed Hub


Rytom71487:
Hi, 

I'm encountering a TypeLoadException when I use a strongly typed hub. My interface is:

    public interface IClientCallback
    {
        void callback<T>(T msg, string eventType);
        void test(string msg, string eventType); 
    }

All the methods are inside a single interface, and the interface does not inherit from any other interface. 

My Hub class is: 

    public class ServiceHub : Hub<IClientCallback>
    {
        public static readonly string ServiceHubName = "ServiceHub";

        public void Register(string name, string eventType)
        {
            Clients.All.test("hello", "world");
        }
    }

When I use my client app to invoke the Register method on the hub, on the Hub application I receive an exception when it is at Clients.All.test(...): 

Method 'callback' in type 'Microsoft.AspNet.SignalR.Hubs.TypedClientBuilder.IClientCallbackImpl' from assembly 'Microsoft.AspNet.SignalR.Hubs.TypedClientBuilder, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.

I haven't been able to narrow down what exactly is causing this exception to be thrown. A little help or suggesting would be greatly appreciated.

moozzyk:
Fixed in 41555309f54184905fe34c7197aa274155faed8c and fbbafd30456a6597b44d4d9b20fac4a3ba8eecfd.

PR #3595, opened 2015-11-14
DefineGenericParameters for type parameters

jhmt:
This is fix for the issue of #3594.

The `TypeLoadException` is thrown while executing `TypeBuilder.CreateType()` in `TypedClientBuilder.GenerateInterfaceImplementation` method.

https://github.com/SignalR/SignalR/blob/3a219c101587333d562f04f30903d50ad9773e1c/src/Microsoft.AspNet.SignalR.Core/Hubs/TypedClientBuilder.cs#L63

I have added `MethodBuilder.DefineGenericParameters` call in BuildMethod method.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Hubs/TypedClientBuilder.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Hubs/TypedClientBuilder.cs
@@ -123,6 +123,11 @@ namespace Microsoft.AspNet.SignalR.Hubs
 
             methodBuilder.SetReturnType(interfaceMethodInfo.ReturnType);
             methodBuilder.SetParameters(paramTypes);
+            // Sets the number of generic type parameters
+            foreach (var generic in paramTypes.Where(param => param.IsGenericParameter))
+            {
+                methodBuilder.DefineGenericParameters(generic.Name);
+            }
 
             ILGenerator generator = methodBuilder.GetILGenerator();
 


src/Microsoft.AspNet.SignalR.Core/Hubs/TypedClientBuilder.cs
```diff
@@ -123,6 +123,11 @@ private static void BuildMethod(TypeBuilder type, MethodInfo interfaceMethodInfo
 
             methodBuilder.SetReturnType(interfaceMethodInfo.ReturnType);
             methodBuilder.SetParameters(paramTypes);
+            // Sets the number of generic type parameters
+            foreach (var generic in paramTypes.Where(param => param.IsGenericParameter))
+            {
+                methodBuilder.DefineGenericParameters(generic.Name);
```
moozzyk: This won't work if the method takes multiple parameters. As per MSDN - . "Calling this method a second time causes an InvalidOperationException.". DefineGenericParameters takes param so the code will need to be changed to pass all the types at once.

moozzyk: Also I think there should not be duplicate names.

moozzyk:
PR #3611 contains this change rebased on the dev branch with a follow-up commit that which a test and support for multiple generic types for a method.

moozzyk:
Merged as 41555309f54184905fe34c7197aa274155faed8c with some follow-up in fbbafd30456a6597b44d4d9b20fac4a3ba8eecfd.
Thanks for your contribution!

PR #3599, opened 2015-11-23
Fix WebSocket connections leak

llooz:
Errors occurring during the initialization of a WebSocket weren’t
notified to the WebSocketTransport that was leaking by returning
IsAlive = true forever

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Owin/WebSockets/DefaultWebSocketHandler.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Owin/WebSockets/DefaultWebSocketHandler.cs
@@ -12,6 +12,9 @@ namespace Microsoft.AspNet.SignalR.WebSockets
         private readonly IWebSocket _webSocket;
         private volatile bool _closed;
 
+        private Action _onClose;
+        private Action<Exception> _onError;
+
         internal ArraySegment<byte> NextMessageToSend { get; private set; }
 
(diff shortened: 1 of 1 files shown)


davidfowl:
When does this happen?

Idhor:
onClose/onError listeners aren't set immediately when using the websocket so if an error occurs during the initialization the transport will be considered alive forever even if the real websocket is closed and disposed
I had this issue with 50 connections leak per day 
If i'm not wrong this is already fixed on signalr for asp.net 5

erichexter:
Bump.  Is this thread still alive ?

Joe-U-Questionmark:
Bump. I would be interested seeing this fix being accepted.

MrDark:
Can confirm this commit fixed our issue where the amount of connections went from a stable 20.000 to 200.000+ causing the server to be unavailable until restarted.

moozzyk:
Looks like a better fix would be to make sure these events are wired up before SignalR starts reading data from the websocket (i.e. before this [line](https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/Owin/OwinWebSocketHandler.cs#L78))

llooz:
Certainly there is a better method to fix this but at that time we needed a rapid fix to fix connection leaks instead of ditching SignalR

moozzyk:
It was fixed by subscribing to events before starting reading data from websocket - cd37171b5e86b777b94e830f2584b896482d2ae1

PR #3604, opened 2015-12-01
LessRed - fixing FxCop warnings after moving to VS2015

moozzyk:


src/Microsoft.AspNet.SignalR.Core/Infrastructure/IMemoryPool.cs
```diff
@@ -6,11 +6,14 @@ namespace Microsoft.AspNet.SignalR.Infrastructure
 {
     public interface IMemoryPool
     {
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1819:PropertiesShouldNotReturnArrays")]
```
moozzyk: @davidfowl - is there any reason why we need this as property on the interface instead of using empty array directly in the code?

src/Microsoft.AspNet.SignalR.Core/Infrastructure/MemoryPoolTextWriter.cs
```diff
@@ -3,10 +3,11 @@
 using System;
 using System.Text;
 using System.IO;
+using System.Globalization;
 
 namespace Microsoft.AspNet.SignalR.Infrastructure
 {
-    public class MemoryPoolTextWriter : TextWriter
+    internal class MemoryPoolTextWriter : TextWriter
```
moozzyk: @davidfowl I made some types internal - was there a specific reason that they were public before (e.g. did we expect people to derive from them and overwrite some functionality)

build/Build.proj
```diff
@@ -94,6 +94,7 @@
     </Projects>
     <Projects Include="$(ProjectRoot)\src\Microsoft.AspNet.SignalR.Client.Store\Microsoft.AspNet.SignalR.Client.Store.csproj">
       <Platform>portable-win81+wpa81</Platform>
+      <RunFxCop>false</RunFxCop>
```
moozzyk: I am following up on this internally. For now it unblocks the build.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/MemoryPoolTextWriter.cs
halter73: I don't think there is a specific reason that so many of SignalR's seemingly internal types are public. @davidfowl (at least at the time) had the philosophy to make all types public unless there was a specific compelling reason not to.

At no point did we ever expect SignalR users to have to include `using  Microsoft.AspNet.SignalR.Infrastructure` in their code. Hopefully making this internal won't break anyone. I doubt it will.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/IMemoryPool.cs
halter73: Of course not :smile:. Other than preferred style maybe.

moozzyk: I am feeling like removing it...

samples/Microsoft.AspNet.SignalR.Samples/Scripts/hubs.js
```diff
@@ -1,5 +1,5 @@
 /*!
- * ASP.NET SignalR JavaScript Library v2.2.0-pre
```
halter73: Does running build.cmd automatically update this? Any idea why this wasn't updated before?

src/Microsoft.AspNet.SignalR.Core/Infrastructure/MemoryPoolTextWriter.cs
moozzyk: This is what I thought - making this internal saves some FxCop suppressions since we probably don't want to check if the value is not null each time we want to write a chunk...

samples/Microsoft.AspNet.SignalR.Samples/Scripts/hubs.js
moozzyk: I must have missed this in a previous commit - I probably didn't stage the file or something to this effect.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/IMemoryPool.cs
davidfowl: Sure

src/Microsoft.AspNet.SignalR.Core/Infrastructure/MemoryPoolTextWriter.cs
davidfowl: That's fine

davidfowl:
:shipit:

PR #3607, opened 2015-12-03
ForYouTheStoreIsClosed

### LINKED ISSUE #3393: Cannot install portable NET client package

moozzyk:
When a project targeted both net45 and wpa81 or win8 we did not have a subset that could satisfy this. The store websocket transport can only be used if an app/lib targets wpa81 and/or win8 since it is using WinRT APIs. If the app/lib targets any other platform in addition to wpa81/win8 the store websocket cannot be used and therefore we must not install it. Adding wpa81/win8 platforms to the folder containing just the portable version of Microsoft.AspNet.SignalR.Client.dll will do the trick - this folder will be selected if the application targets any other platform in addition to wpa81/win8

Fixes #3393

### DIFF
--- a/build/Build.proj
+++ b/build/Build.proj
@@ -81,7 +81,7 @@
     <!-- Only build the following clients if the SDK is installed -->
     <Projects Include="$(ProjectRoot)\src\Microsoft.AspNet.SignalR.Client.Portable\Microsoft.AspNet.SignalR.Client.Portable.csproj">
       <Build Condition="!$(WP8SDKInstalled)">false</Build>
-      <Platform>portable-net45+sl5+netcore45+wp8+wp81</Platform>
+      <Platform>portable-net45+sl50+win+wpa81+wp80</Platform>
     </Projects>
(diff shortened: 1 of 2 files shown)


moozzyk:
@halter73 @anurse - this is basically https://github.com/SignalR/SignalR/pull/3480 but I rebased on latest dev (for some reason all the PRs are against master) and used the target platforms as proposed by @anurse. I tested re-tested with the change and it worked.

halter73:
So long as @anurse says the profile is OK, I'm pretty sure these are the only two files that need changing.

PR #3611, opened 2015-12-09
Jhmt master

moozzyk:
Enabling generic parameters on methods of interfaces used to define strongly named hubs

#3594


tests/Microsoft.AspNet.SignalR.Tests/Server/Hubs/TypedClientBuilderFacts.cs
```diff
@@ -145,6 +154,9 @@ public interface IAmGeneric<T>
         {
             void send(T genericArgument);
             void ping();
+            void send<U>(U u);
+            void send<U, V>(U u, V v);
+            void send<U, V>(U u, V v, U u1);
```
halter73: I'm sure it will pass, but can we add a method that mixes and non-generic parameters?

halter73:
:shipit:

tests/Microsoft.AspNet.SignalR.Tests/Server/Hubs/TypedClientBuilderFacts.cs
moozzyk: Will do

PR #3615, opened 2015-12-17
IndependenceDay

moozzyk:
Creating a dedicated SignalR store client

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client.Store/Microsoft.AspNet.SignalR.Client.Store.csproj
+++ b/src/Microsoft.AspNet.SignalR.Client.Store/Microsoft.AspNet.SignalR.Client.Store.csproj
@@ -9,7 +9,7 @@
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.AspNet.SignalR.Client</RootNamespace>
-    <AssemblyName>Microsoft.AspNet.SignalR.Client.Store</AssemblyName>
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNet.SignalR.Client.Store/Microsoft.AspNet.SignalR.Client.Store.csproj
```diff
@@ -9,7 +9,7 @@
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.AspNet.SignalR.Client</RootNamespace>
-    <AssemblyName>Microsoft.AspNet.SignalR.Client.Store</AssemblyName>
```
halter73: Here's the original PR that added this project: https://github.com/SignalR/SignalR/pull/3028/

I don't see any discussion as to why we went with multiple dlls despite the downside of never automatically choosing the Websocket transport. A quick check of a few follow up PRs also came up empty.

I think making this a single dll is a good change.

halter73:
:shipit:

davidfowl:
:shipit: this is netcore50 right?

moozzyk:
@davidfowl - this is actually portable-win81+wpa81

src/Microsoft.AspNet.SignalR.Client.Store/Microsoft.AspNet.SignalR.Client.Store.csproj
moozzyk: @halter73 I tried digging but could not find anything on this either...

PR #3636, opened 2016-01-28
Update Servicebus to 3.1.2

jeremistadler:


src/Microsoft.AspNet.SignalR.ServiceBus/ServiceBusConnection.cs
```diff
@@ -227,43 +225,19 @@ public void Dispose()
...
-                receiverContext);
-
-                if (result.CompletedSynchronously)
-                {
-                    if (ContinueReceiving(result, receiverContext))
-                    {
-                        goto receive;
-                    }
-                }
+                var task = receiverContext.Receiver.ReceiveBatchAsync(ReceiverContext.ReceiveBatchSize, receiverContext.ReceiveTimeout);
```
davidfowl: This doesn't look very efficient. It's ignoring the synchronous return case.

jeremistadler:
@davidfowl I've changed the implementation to use the ServiceBus package "OnMessage" message pump to avoid handling all that logic. I've been running this in production for 3 weeks without hickups. 

It has been much more reliable and performant than the current official one.

jeremistadler:
@mattscully One easy way to use the new version is to remove the scaleout package and manually add the 4 files to your project. Works like a charm for me :smile:

davidfowl:
There's just too many changes here. Were there all necessary? What behaviors did you change and why? Were old APIs removed in newer versions? If not, I'd suggest doing the minimal change as that will have the highest chance of getting merged, then send a separate PR for improvements.

davidfowl:
You also updated SignalR.Core dependencies as part of this PR. I'm going to close this and recommend that you open up PRs for each change and not lump them into this one.

PR #3647, opened 2016-02-26
LoadTestHarness has incorrectly named "X-UA-Compatibility" meta tag

kb3eua:
Renaming "X-UA-Compatibility" to "X-UA-Compatible"... Although it'd be better to send this as an HTTP response header rather than as a <meta> tag, since sites accessed on the same intranet ignore the meta tag and only respect the response header.


aspnet-hello:

PR #3688, opened 2016-04-28
Adding support for redis cluster as backplane

shykln:
StackExchange.Redis supports redis cluster connection. Adding redis
cluster support by allowing multiple end points to be added to
connection string.


shykln:
@dnfclas It's been signed.

Issue #3714, opened 2016-06-13
Redis Scaleout - Default Db from ConnectionString is ignored


zivillian:
The StackExchange.Redis Connectionstring allows to specify the default db id ([Documentation](https://github.com/StackExchange/StackExchange.Redis/blob/master/Docs/Configuration.md#configuration-options)). This option is currently overwritten by the default value (0) of RedisScaleoutConfiguration.Database

rpannell:
+1  I like to have the database separate for functionality and was shocked that this forced database 0 which I don't want to move around.

moozzyk:
Fixed in 38405a7d5f6944528ee9a70201f30f176e0355e1

PR #3715, opened 2016-06-13
use default Redis Database from ConnectionString

zivillian:
- update StackExchange.Redis to 1.1.603 (to include [c7be31](https://github.com/StackExchange/StackExchange.Redis/commit/c7be31107e8e684ba805253b1fdd66a6017f725b))
- parse defaultDatabase from connectionstring
#3714


moozzyk:
Thanks!

Issue #3710, opened 2016-06-10
jQuery v3.0.0 is unsupported


AntiFreez:
I have updated to jQuery v3.0.0 and caught error:
```
Uncaught TypeError: url.indexOf is not a function
jQuery.fn.load @ jquery.js:9612
(anonymous function) @ jquery.signalR.js:250
(anonymous function) @ jquery.signalR.js:976
```

gerardolima:
Hi, guys, this event is deprecated since [jQuery version 1.8](http://api.jquery.com/load-event/). What do you think about checking other deprecated dependencies from times to times?

FlexHelper:
I dont understand.. 

Oka Jquery don't have .load anymore..? eh

Hmm when i read Document of jquery and Signalr Javascript

I understand the issue

Signalr Use old style


So lets fix the issue


Go in ur Signalr File  ex.:  jquery.signalR-2.2.0.min.js

Replace e.load with $


:) dooone now it will work ;) good luck

PR #3716, opened 2016-06-14
Fix  jQuery v3.0.0 is unsupported

alextkachuk:
- replace deprecated $().load function to $().ready

#3710

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
+++ b/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
@@ -249,7 +249,7 @@
         }
     };
 
-    _pageWindow.load(function () { _pageLoaded = true; });
+    _pageWindow.ready(function () { _pageLoaded = true; });
 
     function validateTransport(requestedTransport, connection) {
         /// <summary>Validates the requested transport by cross checking it with the pre-defined signalR.transports</summary>
--- a/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
+++ b/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
@@ -249,7 +249,7 @@
         }
     };
 
-    _pageWindow.ready(function () { _pageLoaded = true; });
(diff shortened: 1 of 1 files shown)


IanYates:
Confirmed working in my code.  I haven't moved to jQuery 3.0 but was thinking about trying it out soon so you've saved me one headache.  Thanks!

dmethvin:
Note that there can be a big difference between "document ready" which means the HTML document has completely loaded and "window load" which means all resources *inside* the page have loaded as well. The standard way to fix this deprecation is to change `.load(fn)` to `,on("load", fn)` and that's the change I would suggest here.

alextkachuk:
@dmethvin You are right - .load( handler ) is a shortcut for .on( "load", handler ). I changed it.

japalo:
Having the same issue:

```
jquery.js:9612 Uncaught TypeError: url.indexOf is not a function
jQuery.fn.load @ jquery.js:9612
(anonymous function) @ jquery.signalR.js:250
(anonymous function) @ jquery.signalR.js:976
js:18 Uncaught Error: SignalR: SignalR is not loaded. Please ensure jquery.signalR-x.js is referenced before ~/signalr/js.(anonymous function) @ js:18
(anonymous function) @ js:95
```

Is there by any chance possible that you guys update the bower.json dependecy-section from
```
  "dependencies": {
    "jquery": ">=1.6.4"
  }
```
to a working version? As of right now it gets the most recent one i.e. 3.0.0.

Compulsed:
👍

moozzyk:
Thanks for your contribution. I squashed your change and merged into dev as 1feff64b48f611030cfad8827488dbdf6ea91177. It will be included in the 2.2.1 release. 

Next time please run tests and make sure they pass. `.on()` was introduced in version 1.7.0 and the tests were using 1.6.4. While this is a very old version we don't want to make breaking changes like this  in servicing/point releases. The tests also failed when running jquery 3.0.0 due to a breaking change introduced in jquery 3.0.0. I created a follow up commit to sort out these issues.

alextkachuk:
@moozzyk  Good news, thanks!

PR #3727, opened 2016-07-06
Pawelka/jquery

moozzyk:


src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
```diff
@@ -249,7 +249,12 @@
         }
     };
 
-    _pageWindow.load(function () { _pageLoaded = true; });
+    if (typeof _pageWindow.on == "function") {
+        _pageWindow.on("load", function () { _pageLoaded = true; });
```
halter73: I take it that `.on` isn't provided by jQuery 1.6.4, but `.load` isn't provided by 3.0.0? Might deserve a comment.

halter73:
:shipit:

src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
moozzyk: This is correct. I will add a comment.

PR #3735, opened 2016-07-13
Supporting newer version of WindowsAzure.ServiceBus

markheath:
- using extension methods to backfill the missing APM methods `BeginReceiveBatch` and `EndReceiveBatch` in the latest `WindowsAzure.ServiceBus` NuGet packages
- referenced the latest `WindowsAzure.ServiceBus` NuGet Package (3.2.4 at the time of making this), but probably could rely on earlier 3.x versions if that is preferable.
- have tested this in production in Azure, all seems to be working fine.
#3548

### DIFF
--- a/src/Microsoft.AspNet.SignalR.ServiceBus/Infrastructure/ServiceBusTaskExtensions.cs
+++ b/src/Microsoft.AspNet.SignalR.ServiceBus/Infrastructure/ServiceBusTaskExtensions.cs
@@ -1,5 +1,8 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.md in the project root for license information.
 
+using System;
+using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.ServiceBus.Messaging;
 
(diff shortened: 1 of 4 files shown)


njlittle:
+1 for this update

moozzyk:
We created a new package that the depends on Service Bus 3 to not break people who still depend on Service Bus 2 in a servicing release. 0dab7a760ba3d6cbeaf9b219bd913b1ab0f9e9b9..6867865df6c6d04274863a1c9a9743038f485c01

snekbaev:
@moozzyk the new package, is it in Nuget?

moozzyk:
@snekbaev - it will be shipped to NuGet with the 2.2.2 version of SignalR. You can get a pre-release version from here: https://www.myget.org/gallery/aspnetwebstacknightly

snekbaev:
@moozzyk thank you, I'll wait for the rtm :)

christianarg:
Is there an ETA for 2.2.2? We have a kind of "deadlock problem" with all of this:

- On Azure Web Apps there's a known issue when referencing Microsoft.WindowsAzure.Configuration.dll  check David Ebbo answer: http://answers.flyppdevportal.com/MVC/Post/Thread/36a71086-d899-4cbe-ab83-3be722cd7983?category=windowsazurewebsitespreview

- Because Microsoft.Azure.ServiceBus 2.x.x has a dependency against Microsoft.WindowsAzure.ConfigurationManager, which includes that dll that causes startup problems on Azure Web Apps we have to upgrade to Microsoft.Azure.ServiceBus 3.x

- But because of this issue #3735  (Supporting newer version of WindowsAzure.ServiceBus) we can't update to ServiceBus 3

- Which means we are stuck with ServiceBus2, which forces us to include Microsoft.WindowsAzure.Configuration.dll which causes startup problems on Azure Web Apps

hence the "deadlock"

So, any eta on 2.2.2?

moozzyk:
We are working on the release. In the meantime if you need to unblock yourself you can try nightly builds from https://www.myget.org/gallery/aspnetwebstacknightly

moozzyk:
2.2.2-preview1 is now on NuGet

PR #3737, opened 2016-07-13
move UWP Client JSON.Net dep back to 6.0.4

### LINKED ISSUE #3736: Mismatch between package dependency and assembly reference in UWP Client
The UWP Client DLL has a reference to Newtonsoft.Json 7.0.0, but the Microsoft.AspNet.SignalR.Client nuspec references 6.0.4, resulting in Assembly Load failures.

anurse:
Fixes #3736 

Verified locally that this works with the UWP client

/cc @muratg @BrennanConroy @moozzyk


moozzyk:
:ship::it:

PR #3748, opened 2016-07-25
Remove -pre tag from JS 2.2.1-pre

sickboy:


moozzyk:
the `-pre` suffix is being stripped during build for release builds. We don't do changes in master branch.

PR #3794, opened 2016-10-21
Prevent table recreation on deployment

classalaeprod:
Fix SignalR so that it doesn't try to modify the schema if the schema is already current.


src/Microsoft.AspNet.SignalR.SqlServer/install.sql
```diff
@@ -88,7 +88,7 @@ PRINT 'Current SignalR schema version: ' + CASE @CURRENT_SCHEMA_VERSION WHEN NUL
 -- Install tables, etc.
 IF @CURRENT_SCHEMA_VERSION IS NULL OR @CURRENT_SCHEMA_VERSION <= @TARGET_SCHEMA_VERSION
 	BEGIN
-		IF @CURRENT_SCHEMA_VERSION IS NULL OR @CURRENT_SCHEMA_VERSION = @TARGET_SCHEMA_VERSION
+		IF @CURRENT_SCHEMA_VERSION IS NULL OR @CURRENT_SCHEMA_VERSION < @TARGET_SCHEMA_VERSION
```
moozzyk: What if the number of tables changes?

classalaeprod: Why would you add tables without changing the schema version number?

moozzyk: The number of tables is a configuration parameter for Sql Server based scaleout - https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.SqlServer/SqlScaleoutConfiguration.cs#L36

moozzyk:
The number of tables (a.k.a. streams) for Sql Server scale-out is configurable and the configured number of tables needs to be created regardless of the version of the schema.

Issue #3793, opened 2016-10-20
Possible Memory Leak when attempting to Continuously Reconnect


sjones512:
Hello.

We are using the technique outlined in the [documentation](https://www.asp.net/signalr/overview/guide-to-the-api/handling-connection-lifetime-events#continuousreconnect) to attempt to continuously reconnect to the SignalR server. 

In situations where the client is disconnected, and retrying, for an extended period of time, the client side JS memory usage climbs steadily.

Attempting to investigate the issue, it looks like it has to do with onStart event handlers being leaked.

I'm not sure where the issue is, but was wondering if you all had any insight.

We are seeing the issue with SignalR 2.2.0 and jquery 1.10.2.

Thanks for looking.

![image](https://cloud.githubusercontent.com/assets/1839534/19572457/2c640822-96b7-11e6-8d50-05cfd67715ca.png)

![image](https://cloud.githubusercontent.com/assets/1839534/19572638/f7111632-96b7-11e6-9e50-be3c64af6fb7.png)

bigswede74:
I have submitted a pull request above. I tested this fix over a 72hr time period with the JS client continuously reconnecting every 5 seconds and the memory usage was stable.

moozzyk:
Putting tentatively to vNext milestone.

bigswede74:
Sorry that was the pull request to my local repo, I have submitted the real one here #3804

sbabeal:
I was seeing the same type of behavior however i attributed it to the console.log filling up with error messages.  How can handle the errors so they don't appear in the console.log?  I have tried everything window.error => () => {}, etc.  Nothing works.

[...]

PR #3804, opened 2016-11-07
Unbind the onStart event

bigswede74:
Unbind the onStart event before disconnected is called. This can cause a large client memory leak when the client is continuously reconnecting. Fix for issue #3793.


moozzyk:
Thanks!

Issue #3031, opened 2014-05-12
Make scale-out mapping store size configurable


bp74:
Hi,

I'm running a SignalR Scaleout service with multiple web servers on the front side and a Windows Service on the back side. Both share the same Hub and the Windows Service is generating a lot of broadcast messages to different groups of clients. Everything is working fine, except that the Windows Service is loosing lots of memory over time. After the start of the service it need 100MB and after a week or so it is over 1000 MB.

I did a heap snapshot after 2 days an it shows me that all the memory is used here (i hope this is of any use for you):

```
<root> :     260 MB    (100,00%)
 Other, Pinning <root>->Other, Pinning->(System.Object []):  258 MB    (99,50%)
  System.Object [] Other, Pinning->System.Object []->(System.RuntimeType,System.String,System.WeakReference [],...):     247 MB    (95,20%)  (1 object,   32 kB    (0,01%))
   System.Lazy<T> System.Object []->System.Lazy<T>->(System.Func<T>,System.Lazy<T>.Boxed):   241 MB    (92,62%)  (6 objects,  240 bytes (0,00%))
    System.Lazy<T>.Boxed System.Lazy<T>->System.Lazy<T>.Boxed->(Microsoft.AspNet.SignalR.DefaultDependencyResolver):     241 MB    (92,62%)  (1 object,   24 bytes (0,00%))
     Microsoft.AspNet.SignalR.DefaultDependencyResolver System.Lazy<T>.Boxed->Microsoft.AspNet.SignalR.DefaultDependencyResolver->(System.Collections.Generic.Dictionary<T,U>):  241 MB    (92,62%)  (1 object,   40 bytes (0,00%))
      System.Collections.Generic.HashSet<T> Microsoft.AspNet.SignalR.DefaultDependencyResolver->System.Collections.Generic.HashSet<T>->(System.Collections.Generic.ObjectEqualityComparer<T>,System.Int32 []):   241 MB    (92,61%)  (1 object,   64 bytes (0,00%))
       System.Collections.Generic.HashSet<T>.Slot [] System.Collections.Generic.HashSet<T>->System.Collections.Generic.HashSet<T>.Slot []->(Microsoft.AspNet.SignalR.ServiceBus.ServiceBusMessageBus):   241 MB    (92,61%)  (1 object,   72 bytes (0,00%))
[...]

PR #3805, opened 2016-11-09
Enabling configuring max scaleout mappings

moozzyk:


src/Microsoft.AspNet.SignalR.Core/Configuration/IConfigurationManager.cs
```diff
@@ -52,10 +52,16 @@ public interface IConfigurationManager
...
         /// The default value is 0.
         /// </summary>
         [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "long", Justification = "Longpolling is a well known term")]
         TimeSpan LongPollDelay { get; set; }
+
+        /// <summary>
+        /// Gets or sets the maximum number of scaleout mappings per scaleout stream stored on the server.
+        /// The default value is 1000000
+        /// </summary>
+        int? MaxScaleoutMappingsPerStream { get; set; }
```
halter73: Why nullable? I would initialize it to the default value of 1000000 in DefaultConfigurationManager like we do for DisconnectTimeout and similar.

moozzyk: Can do that as well - was following the pattern we used for MaxIncomingWebSocketMessageSize setting...

moozzyk: @halter73 - actually one more reason to use nullable is that I need to add a number of new ctors taking this setting. Since the types are public I keep the old ctors for backwards compatibility and pass null. I initialize to the default value only in one place. If the setting on the IConfigurationManager was not nullable I would have to initialize it to 1000000 in at least two places (DefaultConfigurationManager and ScaleoutMappingStore). In this case I would still keep the nullable parameter on ScaleoutStreamManager ctor. If you wanted to make this param non-nullable then I would have to initialize to the default value there as well. Nullable allows me to initialize only in one place.

moozzyk:
:up::date:

moozzyk:
https://github.com/SignalR/SignalR/issues/3031

src/Microsoft.AspNet.SignalR.Core/Messaging/ScaleoutMappingStore.cs
```diff
@@ -4,26 +4,30 @@
...
 {
     public class ScaleoutMappingStore
     {
-        private const int MaxMessages = 1000000;
-
         private ScaleoutStore _store;
+        private readonly uint _maxMessages;
 
         public ScaleoutMappingStore()
+            : this(null)
```
moozzyk: Yes. I think it would be better. Will fix.

moozzyk:
:up::date:

halter73:
:shipit:

src/Microsoft.AspNet.SignalR.Core/Configuration/DefaultConfigurationManager.cs
```diff
@@ -14,12 +14,15 @@ public class DefaultConfigurationManager : IConfigurationManager
         // if _minimumKeepAlivesPerDisconnectTimeout != 3, update the ArguementOutOfRanceExceptionMessage below
         private const int _minimumKeepAlivesPerDisconnectTimeout = 3;
 
+        internal const int DefaultMaxScaleoutMappingsPerStream = 1000000;
```
davidfowl: Should we change the default?

moozzyk: I left it for "compatibility". I am not sure if we know what the "correct" value should be (although it seems that 50000/100000 could be sufficient)

halter73: I don't think we should worry too much about compatibility here. Especially now that it's configurable.

Before we were extremely conservative with the max scaleout mappings in terms of making sure we wouldn't lose messages during reconnects. I remember the conversations back then and they basically amounted to "memory is cheap." As we've discovered, in many cases it's not.

moozzyk: so whats the new number? 65535?

halter73: Sounds good to me.

The basic idea is that it should be greater than the total number of messages sent through the backplane to anyone (not just the reconnecting client) in the time it takes for a client to reconnect should be less than the new number. I think that anything over 1000 is probably good enough for most cases. 65535 is probably safer, and now it's configurable, so if there's problems it can always be adjusted.

PR #3806, opened 2016-11-10
OutOfBodyExperience - removing IFrame from body

### LINKED ISSUE #2795: Object Expected Error Line : 1285  onInitialized(); When trying to use signalr(forevrframe) with jQuery UI Dialog
Hi,

When we try to use jQuery UI Dialog with Signalr we are getting the error Object Expected Error Line : 1285  onInitialized();

This happens with >= IE9 version (foreverframes)

To reproduce the issue download the sample code from http://sdrv.ms/1aAqYRD and host it in IIS.

1) Make sure websocket is diabled in IIS.
2) Run the sample in IE version >= 9 with debug mode enabled and click on "click this". 
3) You will get the error Object Expected on line no 1285 in SignalR js file.

Is there a way to solve this jscript signalr error.

Thank You.

jp9506:
See commit c9200820823395b1b31d3b800938568e496ca2c5 by moozzyk on Aug 13, 2014

foreverFrame was moved to documentElement however stop was still looking for it in the body.  End result was connection was disposed but frame survived causing null reference errors to be thrown.

Fixes #2795

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.transports.foreverFrame.js
+++ b/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.transports.foreverFrame.js
@@ -212,8 +212,8 @@
                 }
 
                 // Ensure the iframe is where we left it
-                if (connection.frame.parentNode === window.document.body) {
-                    window.document.body.removeChild(connection.frame);
+                if (connection.frame.parentNode === window.document.documentElement) {
+                    window.document.documentElement.removeChild(connection.frame);
                 }
 
                 delete transportLogic.foreverFrame.connections[connection.frameId];


moozzyk:
This is a good find. Thanks!

moozzyk:
Can you add a test for this (or update the existing test)?

jp9506:
Unfortunately I am not familiar enough with github/test environment/etc. to be able to do that.

Bug was found in our production environment which is implementing a method similar to the one found here:
https://www.asp.net/signalr/overview/guide-to-the-api/handling-connection-lifetime-events#continuousreconnect

I was able to replicate the problem by disabling my LAN connection.  Once the connection was re-enabled, the reconnect code worked but the javascript console was logging an error every 10 sec on the frame that wasn't removed.  Given the unreliable connectivity of our user-base this spiraled out of control fairly quickly.

moozzyk:
OK. I added the test. Thanks!

PR #3827, opened 2016-12-11
Added EXE support (for OWIN self-hosted SignalR hubs)

jcoutch:
The Signalr.exe utility only scanned for DLL's.  In the case of a self-hosted OWIN process, the hubs are generally part of the EXE.  Added a /exe flag that will search EXE's first, followed by DLL's.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Utils/Commands/GenerateHubProxyCommand.cs
+++ b/src/Microsoft.AspNet.SignalR.Utils/Commands/GenerateHubProxyCommand.cs
@@ -38,8 +38,9 @@ namespace Microsoft.AspNet.SignalR.Utils
             string path = null;
             string outputPath = null;
             string url = null;
+            var includeExecutables = false;
 
-            ParseArguments(args, out url, out path, out outputPath);
+            ParseArguments(args, out url, out path, out outputPath, out includeExecutables);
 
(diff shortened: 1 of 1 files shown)


davidfowl:
This doesn't need to be a flag really.

jcoutch:
Sorry I didn't respond right away.  Primary reason I tossed it behind a flag was so it didn't affect existing behavior.  I can remove the flag if it's not needed.

jcoutch:
@davidfowl - I removed the /exe flag, and made it the default behavior.

src/Microsoft.AspNet.SignalR.Utils/Commands/GenerateHubProxyCommand.cs
```diff
@@ -61,7 +61,9 @@ private void OutputHubs(string path, string url, string outputPath)
             path = path ?? Directory.GetCurrentDirectory();
             url = url ?? "/signalr";
 
-            var assemblies = Directory.GetFiles(path, "*.dll", SearchOption.AllDirectories);
+            var assemblies = Directory.GetFiles(path, "*.exe", SearchOption.AllDirectories)
```
moozzyk: Should be `Concat` instead of `Union` since the items are already unique.

Issue #3829, opened 2016-12-14
The client should reuse HttpClient for negotiate and start requests


moozzyk:
### Expected  behavior
SignalR client reuses the same HttpClient for negotiate and start requests. All SSE and LongPolling connect/poll requests should use the same httpClient for long running requests

### Actual behavior
SignalR client calls IHttpClient.Initialize before sending negotiate and start requests. The DefaultHttpClient creates httpClient in the Initialize method. Since start can be sent only after negotiate initializing the IHttpClient when starting the request should not be needed. Currently it resets the clients that were used to send the negotiate (which is short running) and connect (which is long running) requests. This may result in breaking sticky sessions in load balancers.

moozzyk:
Fixed in 0b362f6

PR #3830, opened 2016-12-14
Removing extraneous call to IHttpClient.Initialize

moozzyk:
An extraneous call to IHttpClient.Initialize when sending the start requests resets httpClients in the DefaultHttpClient. This may break load balancers and prevent reusing HTTPS connections (SSL negotation will happen twice - once for negotiate and once for start)

More: #3829

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Transports/TransportHelper.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Transports/TransportHelper.cs
@@ -57,8 +57,6 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
 
             var startUrl = UrlBuilder.BuildStart(connection, transport, connectionData);
 
-            httpClient.Initialize(connection);
-
             return httpClient.Get(startUrl, connection.PrepareRequest, isLongRunning: false)
                             .Then(response => response.ReadAsString());
         }


src/Microsoft.AspNet.SignalR.Client/Transports/TransportHelper.cs
```diff
@@ -57,8 +57,6 @@ public virtual Task<string> GetStartResponse(IHttpClient httpClient, IConnection
 
             var startUrl = UrlBuilder.BuildStart(connection, transport, connectionData);
 
-            httpClient.Initialize(connection);
```
halter73: It seems kinda fragile depending on GetNegotiationResponse to be called immediately before GetStartResponse on the same IHttpClient. I don't think either of the Get* methods should call Initialize if they both can't.

We should call Initialize in Connection.Start(IHttpClient httpClient) instead.

moozzyk: To send the start request you need the connectionToken token you receive in the response to the negotiate request. Given that, this logic is hardcoded in the client and not something a user can control I think we can rely on the fact that negotiate will be called right before start. 
Initializing the httpClient when sending the start does not fully fix the issue since in this case the negotiate request will use a different http client than start request potentially resulting in sending requests to different servers and forcing separate https handshakes.

halter73: I understand the desire to remove the double initialize given the .NET 4.5 version of DefaultHttpClient.

It's more difficult than I first realized to move the call to Initialize outside of TransportHelper. So I would just :shipit:

moozzyk: Yeah, it would be much bigger change.

PR #3836, opened 2017-01-11
Preventing pipeline from overriding original callstack ...

pasn:
... when handling exception from server-side hub method.

SignalR is currently sending call stack from "await" in HubPipelineModule.BuildIncoming(). This fix changes this behavior, so that call stack from hub method is not overridden.


davidfowl:
Why is this a problem? The async state machine is a reality of using async await.

pasn:
The issue is with the CallStack that may be optionally sent from Hub to Client when exception occurs.

I would like the exception to contain the real stack trace and not the internal one from SignalR.

pasn:
Bump.

aspnet-hello:

PR #3841, opened 2017-01-24
'signalr/negotiate' CustomAuthorizeAttribute conflicts 

msx752:
when we use **CustomAuthorizeAttribute** on our signalr server, anybody who call `signalr/negotiate` without parameters(malicious person) can give access from signalr **without visit** the CustomAuthorizeAttribute

**Examples**
`http://shootr.signalr.net/signalr/negotiate`
**and**
`http://shootr.signalr.net/signalr/negotiate?connectionData=%5B%7B%22name%22%3A%22h%22%7D%5D&clientProtocol=1.3&_=1485297186225`

- firstone can access without visit CustomAuthorizeAttr and their **"ProtocolVersion":"1.2"**
- secondone visits to CustomAuthorizeAttr then maybe it can be declined and their **"ProtocolVersion":"1.3"**

am i wrong?. Can somebody enlighten me.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Hubs/HubDispatcher.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Hubs/HubDispatcher.cs
@@ -141,9 +141,13 @@ namespace Microsoft.AspNet.SignalR.Hubs
                 }
             }
 
-            return base.AuthorizeRequest(request);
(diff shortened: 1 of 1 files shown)


moozzyk:
connectionData contains hubs the client wants to access (after url decoding connectionData from this url: http://shootr.signalr.net/signalr/negotiate?connectionData=%5B%7B%22name%22%3A%22h%22%7D%5D&clientProtocol=1.3&_=1485297186225 you will get: `[{"name":"h"}]`). In the second example there is no connectionData so there is no hubs so there is  nothing to authorize the user to.

PR #3845, opened 2017-01-26
Pawelka/service bus

moozzyk:
Support for Service Bus 3

Fixes https://github.com/SignalR/SignalR/issues/3548


nuspecs/Microsoft.AspNet.SignalR.ServiceBus3.nuspec
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<package>
+  <metadata>
+    <id>Microsoft.AspNet.SignalR.ServiceBus3</id>
```
davidfowl: Wut?

moozzyk: Because we upgrade the dependency from 2 to 3 which has API breaking changes. We can talk about versioning with @muratg and @DamianEdwards offline. The most I care about in this PR is actually code changes.

moozzyk:
There is also an alternative change (PR) created by community: https://github.com/SignalR/SignalR/pull/3735

nuspecs/Microsoft.AspNet.SignalR.ServiceBus3.nuspec
DamianEdwards: We can talk about this again. It's messy any way we go. We didn't like the idea of revving the entire SignalR package set to 3 because of this, but maybe we should. We need to decide if we're going to update to latest Katana packages too, which when through their own major version bump.

src/Microsoft.AspNet.SignalR.ServiceBus3/Infrastructure/ServiceBusTaskExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.ServiceBus.Messaging;
+
+namespace Microsoft.AspNet.SignalR.ServiceBus.Infrastructure
+{
+    public static class ServiceBusTaskExtensions
```
davidfowl: What's the purpose of this extension method?

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
```diff
@@ -0,0 +1,328 @@
...
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Exceptions are handled through the error handler callback")]
+        private void ProcessMessages(ReceiverContext receiverContext)
+        {
+            receiverContext.Receiver.ReceiveBatchAsync(ReceiverContext.ReceiveBatchSize, receiverContext.ReceiveTimeout)
```
davidfowl: I would just switch to async/await

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
```diff
@@ -0,0 +1,328 @@
...
+                .ContinueWith(t =>
+                {
+                    ContinueReceiving(t, receiverContext);
+                });
+        }
+
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Exceptions are handled through the error handler callback")]
+        private void ContinueReceiving(Task<IEnumerable<BrokeredMessage>> receiveTask, ReceiverContext receiverContext)
```
davidfowl: async/awaitify

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
moozzyk: @DamianEdwards recommended against async/await to avoid issues related to old Asp.NET threading model.

src/Microsoft.AspNet.SignalR.ServiceBus3/Infrastructure/ServiceBusTaskExtensions.cs
moozzyk: This an artifact of sharing code. This project has its own implementation of only ServiceBusConnection and SendAsync and all other code is from the other project. SendAsync is invoked from [ServiceBusConnectionContext](https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.ServiceBus/ServiceBusConnectionContext.cs#L82) which is shared code. In the other project it is an [extension method](https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.ServiceBus/Infrastructure/ServiceBusTaskExtensions.cs) so I just did the same here.

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
davidfowl: This is a random background thread. You can also just task.run or use this beauty https://github.com/aspnet/SignalR/blob/eafbe741609981cdcf6b6f28d45cb2f50126750d/src/Microsoft.AspNetCore.Sockets/Internal/AwaitableThreadPool.cs. Usage here https://github.com/aspnet/SignalR/blob/eafbe741609981cdcf6b6f28d45cb2f50126750d/src/Microsoft.AspNetCore.Sockets/HttpConnectionDispatcher.cs#L291

moozzyk: OK. I will look into it. I am all for async'ifying (is it even a word?) this.

moozzyk:
:up::date:

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
```diff
@@ -0,0 +1,326 @@
...
+                catch (MessagingEntityNotFoundException ex)
+                {
+                    try
+                    {
+                        await receiverContext.Receiver.CloseAsync();
+                    }
+                    catch { }
+
+                    receiverContext.OnError(ex);
+                    var _ = TryCreateSubscription(receiverContext);
```
davidfowl: call this ignore because _ looks bizzare

moozzyk: I thought `_` is a quite common convention (go, Swift and EF uses it) but I can change.

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
```diff
@@ -0,0 +1,326 @@
...
+                string subscriptionEntityPath = SubscriptionClient.FormatSubscriptionPath(topicName, subscriptionName);
+                MessageReceiver receiver = _factory.CreateMessageReceiver(subscriptionEntityPath, ReceiveMode.ReceiveAndDelete);
+
+                _trace.TraceInformation("Creation of a message receive for subscription entity path {0} in the service bus completed successfully.", subscriptionEntityPath);
+
+                connectionContext.SetSubscriptionContext(new SubscriptionContext(topicName, subscriptionName, receiver), topicIndex);
+
+                var receiverContext = new ReceiverContext(topicIndex, receiver, connectionContext);
+
+                Task.Run(() => ProcessMessages(receiverContext));
```
davidfowl: Should dispose wait on this task?

moozzyk: We didn't before so I did not look into it. I am not sure what would be the consequences of synchronously blocking here.  I guess it would also have to have time out in which case if the timeout expires we are in the same boat - Dispose() finished task is not finished. But I will look into it.

davidfowl: Yea, it would have to have a timeout. I'm mostly wondering if we care about making sure things are winding down and not hanging around with a rooted task that we don't track.

moozzyk: The problem is that even with .Wait this won't be guaranteed - it still can time out.

moozzyk:
:up::date:
@davidfowl - I added timeout in Dispose(). I have mixed feelings though - Dispose was never called with selfhost. With IISExpress it did get called but as soon as `_factory.Close()` is executed a ThreadAbortException is being thrown ServiceBus and and `WaitAll` is never reached....

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
```diff
@@ -169,7 +173,8 @@ private void CreateSubscription(ServiceBusConnectionContext connectionContext, i
 
                 var receiverContext = new ReceiverContext(topicIndex, receiver, connectionContext);
 
-                Task.Run(() => ProcessMessages(receiverContext));
+                var t = Task.Run(() => ProcessMessages(receiverContext));
+                _processMessagesTasks.Add(t);
```
davidfowl: You should remove the task on completion

moozzyk:
Needed to switch to ConcurrentDictionary to be able to remove tasks.

davidfowl:
> @davidfowl - I added timeout in Dispose(). I have mixed feelings though - Dispose was never called with selfhost. With IISExpress it did get called but as soon as _factory.Close() is executed a ThreadAbortException is being thrown ServiceBus and and WaitAll is never reached....

Thread abort!? Ok we don't need to wait for it then.

moozzyk:
I thought it would when the IDisposable returned by WebApp.Start is disposed but my breakpoint in Dispose was never hit.

davidfowl:
Ok lets close on this today. The logic looks fine. Add a log on timeout if WaitAll returns false.

src/Microsoft.AspNet.SignalR.ServiceBus3/ServiceBusConnection.cs
```diff
@@ -0,0 +1,340 @@
...
+            if (disposing)
+            {
+                // Close the factory
+                if (_factory != null)
+                {
+                    _factory.Close();
+                }
+
+                _cts.Cancel();
+                Task.WaitAll(_processMessagesTasks.Keys.ToArray(), DefaultReadTimeout);
```
davidfowl: What does this have to do with the ReadTimeout? Is that when the task will unwind?

moozzyk: This is the longest timeout we are using in ServiceBusConnection so this should give a chance to finish all the tasks cleanly before giving up.

moozzyk:
Added writing a traces in case of timeouts.

PR #3846, opened 2017-01-27
Making sure minified script contains trailing semicolon

### LINKED ISSUE #3516: Signalr bower jquery.signalR.min.js missing semicolon in the end
If this JS file is present in a minified file which contains several Javascript files, can cause Javascript problems on the first line of the next included file.

moozzyk:
From AjaxMinifier.exe -?
 -term > ensure output ends in semicolon or other valid terminator.

 Fixes #3516


BrennanConroy:
:shipit:

Issue #3767, opened 2016-08-31
Redis script uses textual comparison


moozzyk:
_Lua will automatically convert string and number types to the correct format to perform calculations. (...) A notable exception: comparison operators (== ~= < > <= >=) do not coerce their arguments._

Consider:

```
127.0.0.1:6379> SET test 50
OK
127.0.0.1:6379> EVAL "local newvalue = tonumber(redis.call('GET', KEYS[1])) if newvalue < tonumber(ARGV[1]) then return redis.call('SET', KEYS[1], ARGV[1]) else return nil end" 1 test 6
(nil)
```

vs. the script in the current form:

```
127.0.0.1:6379> SET test 50
OK
127.0.0.1:6379> EVAL "local newvalue = redis.call('GET', KEYS[1]) if newvalue < ARGV[1] then return redis.call('SET', KEYS[1], ARGV[1]) else return nil end" 1 test 6
OK
```

This script runs only when reconnecting to Redis.

PR #3847, opened 2017-01-27
Converting strings to numbers for comparison in lua script

moozzyk:
_Lua will automatically convert string and number types to the correct format to perform calculations. (...) A notable exception: comparison operators (== ~= < > <= >=) do not coerce their arguments._

Fixes: #3767


moozzyk:
Manually verified that the values are updated.

PR #3849, opened 2017-01-27
Preventing from running connection event handlers twice

### LINKED ISSUE #3729: Race in Redis scaleout when the connection is restored
Looks like when a connection to Redis is restored the event handler can be called more than 1 time. 

image

We would then reset the key in Redis multiple times.

moozzyk:
Because SignalR executes Redis scripts and uses Redis subscryptions it receives ConnectionFailed/ConnectionRestored events twice - once for ConnectionType.Interactive and once for ConnectionType.Subscription. This change prevents from running a subsequent event if the previous event completed successfully.

Fixes #3729

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
+++ b/src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
@@ -27,6 +27,7 @@ namespace Microsoft.AspNet.SignalR.Redis
         private string _connectionString;
         private int _state;
         private readonly object _callbackLock = new object();
+        private readonly SemaphoreSlim _redisConnectionEventLock = new SemaphoreSlim(1, 1);
 
         public RedisMessageBus(IDependencyResolver resolver, RedisScaleoutConfiguration configuration, IRedisConnection connection)
             : this(resolver, configuration, connection, true)
@@ -106,6 +107,8 @@ namespace Microsoft.AspNet.SignalR.Redis
                     default:
                         break;
                 }
+
+                _redisConnectionEventLock.Dispose();
             }
 
             base.Dispose(disposing);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
```diff
@@ -27,6 +27,7 @@ public class RedisMessageBus : ScaleoutMessageBus
         private string _connectionString;
         private int _state;
         private readonly object _callbackLock = new object();
+        private readonly SemaphoreSlim _redisConnectionEventLock = new SemaphoreSlim(1, 1);
```
davidfowl: Why not just a lock?

moozzyk: because https://github.com/SignalR/SignalR/pull/3849/files/45b4089f1d23dc101284a7d7b945f246384baadc#diff-193a338ff4a86e8ec19aa6cfa2954768R189 (await expressions cannot be used in a synchronous function)

davidfowl: Ahh, why not WaitAsync then?

moozzyk: I didn't feel it really makes a difference given that the other event handler always finishes very quickly and the collision is very unlikely.

moozzyk:
@davidfowl - is it good to go or you feel strongly about `WaitAsync` (in which case should we also `WaitAsync` in `OnConnectionFailed` and make it `async void` as well)?

src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
davidfowl: Part of me is wondering now if we should switch to interlocked all over or lock all over. Mixing the 2 things seems strange.

moozzyk: Wouldn't it mean that we would have to block on executing the Redis script though?

PR #3853, opened 2017-01-31
add netstd to client nuspec to support xamarinmac20

tofutim:
Although it should be possible, SignalR cannot currently be added to xamarinmac20 (Xamarin Mac Mobile Framework) because the package targets 'portable-net45+sl50+win+wpa81+wp80' which the nuget installer in Xamarin Studio does not recognize as being valid. Targeting netstandard1.1 however, will allow it to work.


moozzyk:
The question is if this is the right version to add. 'portable-net45+sl50+win+wpa81+wp80' is the basic portable client that does not have the WebSockets transport implemented. I can see that ClientWebSocket is available so it should be possible to use a different client version that includes the WebSockets transport.

tofutim:
That is even better. Looking forward to seeing the netstandard1.1 version then.

tofutim:
Is ClientWebSocket only in netstandard2.0? https://apisof.net/catalog/System.Net.WebSockets.ClientWebSocket

PR #3855, opened 2017-01-31
initial Mac client

tofutim:
I was hoping for Shared library but ended up making links to everything in Client45. It compiles, but I haven't tested yet. Theoretically it would enable WebSockets for Xamarin Mac. Similar strategy could be used to make iOS and Android versions.


tofutim:
I'm building from Xamarin.Mac from the Mono project. Checking in Visual Studio (Windows), I can compile Android, iOS, tvOS. I cannot load watchOS. And Mac loads but there is a multiple similar errors in building like

> Severity	Code	Description	Project	File	Line	Suppression State
> Error	CS0012	The type 'Attribute' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.	Microsoft.AspNet.SignalR.Client.Mac	D:\Loqu8\Apps\SignalR\src\Microsoft.AspNet.SignalR.Client\Hubs\HubInvocation.cs	13	Active
> 

which seem like Facade issues.

tofutim:
the projects are all compiling, however, build.cmd yields some errors and warnings I haven't been able to figure out. Any insights much appreciated.

moozzyk:
I don't think we will be able to take this. netstandard is the way to go.

tofutim:
you mean netstandard 2.0?

moozzyk:
The lowest that is possible which seems to be netstandard1.3 (1.2 and below do not have `Timer`)

tofutim:
Note that with this approach Sockets would be supported in the Xamarin platforms.

moozzyk:
We decided to cross compile the client for netstandard1.3 which will make the client work on xamarinmac20. ff815cf392e47b331b423e900f2dad72f506ff4e

PR #3862, opened 2017-02-09
fix #3093 by cleaning up Redis connection state

### LINKED ISSUE #3093: With Redis scale-out, once Send() method runs into an error, the stream will never be opened again until Redis is restarted
With Redis scale-out, once Send() method runs into an error,  if the instance doesn't receive any message, then in the instance the stream will never be opened again until Redis is restarted.  No connections in the instance can send a message. Instead they return the same error.  Restarting Redis fixes the issue.
##### Repro steps:

1). Use AspNet SignalR sample, update it to use Redis scale-out,  just let one web app use the Redis, start the Redis
2). Since the issue not easy to repro, I add below code in the RedisMessageBus.Send() method to throw when message contains “throw”:

``` csharp
        protected override Task Send(int streamIndex, IList<Message> messages)
        {
            var message = messages[0].GetString();
            if (message.Contains("throw"))
            {
                throw new Exception("Verify Throwing an exception from Send()");
            } 
            . . . . . . 
        }
```

3). Browse to Hubs/HubConnectionAPI/Default.aspx page 
4). Input "throw" in "To Everybody" textbox, Click button "Broadcast".  This will cause Send method throw.
5). Input "aaaa" in "To Everybody" textbox, Click button "Broadcast".  Observe " Error: Verify Throwing an exception from Send()"
6). Wait for 1 minute or more, click "Broadcast" button again, Repeat this step several times.
##### Expected result:

Like SB scale-out /SqlServer scale-out, "Broadcast" can send the messages successfully.
##### Actual result:

"Error: Verify Throwing an exception from Send()" was always thrown for sending the messages which should not cause Send method throw in step (6).

anurse:
also: add some tracing to redis message bus code

fixes #3093

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
+++ b/src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
@@ -17,7 +17,11 @@ namespace Microsoft.AspNet.SignalR.Redis
 
         public async Task ConnectAsync(string connectionString, TraceSource trace)
         {
-            _connection = await ConnectionMultiplexer.ConnectAsync(connectionString);
+            _connection = await ConnectionMultiplexer.ConnectAsync(connectionString, new TraceTextWriter("ConnectionMultiplexer: ", trace));
+            if(!_connection.IsConnected)
+            {
+                throw new InvalidOperationException("Failed to connect to Redis");
+            }
 
             _connection.ConnectionFailed += OnConnectionFailed;
             _connection.ConnectionRestored += OnConnectionRestored;
@@ -31,6 +35,7 @@ namespace Microsoft.AspNet.SignalR.Redis
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1026:DefaultParametersShouldNotBeUsed")]
         public void Close(string key, bool allowCommandsToComplete = true)
         {
+            _trace.TraceInformation("Closing key: " + key);
             if (_redisSubscriber != null)
             {
                 _redisSubscriber.Unsubscribe(key);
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
```diff
@@ -17,7 +17,11 @@ public class RedisConnection : IRedisConnection
 
         public async Task ConnectAsync(string connectionString, TraceSource trace)
         {
-            _connection = await ConnectionMultiplexer.ConnectAsync(connectionString);
+            _connection = await ConnectionMultiplexer.ConnectAsync(connectionString, new TraceTextWriter("ConnectionMultiplexer: ", trace));
+            if(!_connection.IsConnected)
+            {
```
moozzyk: Should we Dispose _connection here (and possible set to null?)

src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
```diff
@@ -17,7 +17,11 @@ public class RedisConnection : IRedisConnection
...
+                throw new InvalidOperationException("Failed to connect to Redis");
+            }
 
             _connection.ConnectionFailed += OnConnectionFailed;
```
moozzyk: not related to this change but we seem to be subscribing to these events twice...

src/Microsoft.AspNet.SignalR.Redis/RedisConnection.cs
anurse: Well, when it does "re-connect", the object actually becomes active again, so maybe not.

PR #3864, opened 2017-02-10
Subscribing to events before starting processing webSocket messages

moozzyk:
Previously we would subscribe to webSocketHandler events after starting processing webSocket messages this could lead to missing some events. After the change we would subscribe to webSocketHandler events before we start processing webSocket messages.


davidfowl:
@moozzyk was this change originally about a memory leak or missing messages? Either way the change looks good but was just unclear about the original issue.

moozzyk:
It's about missing events - both messages and Close events. If the Close event is missed (i.e. the websocket is closed before we subscribe to the Close event) the transport will never be marked as not alive and hence will never be discarded even though the websocket has been closed.

moozzyk:
Nope. It is also quite hard to test with tests since this is a race. I verified this scenario manually.

PR #3877, opened 2017-02-24
Add an option for the provider to be a publisher only

celeronpm:
I have many applications which write events to the signalR bus. These
modifications allow the provider to toss received messages and forego
the processing of them while focusing on publishing to the redis
instance.


moozzyk:
This is not something we are going to take.

celeronpm:
Ok, may I know a specific reason for this?

I have a dozen servers running on the backplane generating several thousand events a second across the farm (all of which must be digested by each server). This option in the provider was a must for me to be able to continue scaling the farm without spending unnecessary CPU time of each server working though each message. 

Is there a different approach I should be taking to achieve scale with the backplace provider then?

moozzyk:
There are several reasons:
 - I think this change is very risky. I don't know the consequences of cheating the MessageBus and MessageBus was not designed to be cheated. 
 - We have more scaleout providers than just Redis and this introduces a behavior that other providers don't have
 - This feature feels like a one-off thing that does not fit the framework (and I think it can be achieved without changing the framework - see below)
 - It's breaking binary compatibility by changing the public API

If you only want to publish then I think there is no reason to use backplane at all. You can publish messages from outside the hub using the hub context you get from GlobalHost (you can find more details in [this post](https://weblogs.asp.net/ricardoperes/sending-messages-to-signalr-hubs-from-the-outside)). When you have something to publish just invoke your publish method on each server and be done with it. If you want to use Redis for this each server could subscribe to a Redis subscription and then they would be notified when there is a new message to publish and publish the message to connected clients via GlobalHost/HubContext. 

You may also want to take a look at the [stock ticker sample](https://docs.microsoft.com/en-us/aspnet/signalr/overview/getting-started/tutorial-server-broadcast-with-signalr) which pushes notifications to clients periodically. If you want to publish to multiple servers the idea is similar only you need to be able to push to multiple servers and when you have the data. Using Redis subscription models seems to fit the bill.

celeronpm:
Hmm, thank you for the suggestions. 

Basically, I have service workers which do async processing on the side. When they need to, they use the solution I submitted to post a message on the backplane (since they participate in the backplane). The web servers which are doing the actual client signalr connection then read these posted messages and then emit them to the appropriate clients.

I could do a redis subscription or a rabbitmq side queue from the workers to the signalr hosts as you suggested, but I feel that is unnecessary overhead. Additionally, having the async worker, which participates in the backplane, read every message was also overhead (thus my solution).  

Looking at the link your provided about posting from outside the hub, the problem is that in order for the backplane to properly queue the message, it needs to participate in order to have the redis counter (so it knows the new message ID it can use). My solution allows the participation in the backplane without the overhead of unnecessarily reading and emiting the backplane messages to the app.

I understand you reasoning and will continue to maintain this branch for folks who need such a specific solution.

PR #3883, opened 2017-03-05
make LoadCounter method virtual

roman-76:
This change will allow to create custom PerformanceCounterManager with overridden LoadCounter that will create custom performance counters


moozzyk:
You can implement the `IPerformanceCounterManager` interface to provide your own implementation of PerformanceCounterManager.

Issue #3645, opened 2016-02-23
SignalR Client Receives enormous duplicate messages 


karthiik007:
**Environment**
2 WebAPI Server: Windows Server 2012 R2, IIS 8.5, .NET 4.5, SignalR 2.1.1 (client) , SigalR Server 2.1.2 with SqlScaleout

**DBMS:** Microsoft SQL Server 2012 R2 (SP1) - 11.0.3000.0 (X64)

**SqlScaleoutConfiguration:**

string sqlConnectionString = ConfigurationManager.ConnectionStrings["SignalRDB"].ConnectionString;
            GlobalHost.DependencyResolver.UseSqlServer(sqlConnectionString);
            //app.MapSignalR();
            Microsoft.AspNet.SignalR.HubConfiguration huc = new Microsoft.AspNet.SignalR.HubConfiguration();
            huc.EnableDetailedErrors = true;
            app.MapSignalR(huc);

**Description**
Clients receive a lot of duplicate messages. But signalR server has sent the Message only once to all the clients.

when we checked the SignalRDB database logs , there is lot of payload with messages like as follows : 

**{"WaitForAck":false,"Id":"9f7d2c6f-7ffa-4771-9353-7c767403a15b","CommandType":0,"Value":null}$9f7d2c6f-7ffa-4771-9353-7c767403a15bb**

We are not very sure what is that WaitForAck and why its false.Moreover the WaitForAck count ~4000 , but each client has recevied the duplicate message ~20000.

Any clues what could trigger this to client to receive ~20000 messages ?

moozzyk:
Do you know what transport did you use? Also, is there any way to repro this?

karthiik007:
it looks like if we make the SignalR DB very busy we can repro it , but i havent tried reproducing it  ,Let me try and get back if i can repro it ,

Checked the connection object of the client code while debugging and its LongPolling Transport.

moozzyk:
If you can repro this - can you capture traces from the client and from the server. Here is an article describing how to enable tracing http://www.asp.net/signalr/overview/testing-and-debugging/enabling-signalr-tracing

karthiik007:
Sure , let me revert once i repro this.
Many thanks Moozzyk

allenjzhang:
We are also seeing occasionally 10's thousands duplicate messages with Service Bus Scale out. @moozzyk Since we are in production, it is difficult to capture process dump.

moozzyk:
[...]

PR #3887, opened 2017-03-07
Adding additional logging for Scaleout and transports

moozzyk:
Should help figure out #3645

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Messaging/MessageBus.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Messaging/MessageBus.cs
@@ -152,7 +152,7 @@ namespace Microsoft.AspNet.SignalR.Messaging
             if (Topics.TryGetValue(message.Key, out topic))
             {
                 topic.Store.Add(message);
-                ScheduleTopic(topic);
+                ScheduleTopic(message.Key, topic);
             }
 
             Counters.MessageBusMessagesPublishedTotal.Increment();
@@ -258,11 +258,11 @@ namespace Microsoft.AspNet.SignalR.Messaging
             Topic topic;
             if (Topics.TryGetValue(eventKey, out topic))
             {
-                ScheduleTopic(topic);
+                ScheduleTopic(eventKey, topic);
             }
         }
 
-        private void ScheduleTopic(Topic topic)
+        private void ScheduleTopic(string eventKey, Topic topic)
         {
             try
             {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNet.SignalR.Core/Messaging/MessageBus.cs
```diff
@@ -258,16 +258,18 @@ protected void ScheduleEvent(string eventKey)
...
         }
 
-        private void ScheduleTopic(Topic topic)
+        private void ScheduleTopic(string eventKey, Topic topic)
         {
             try
             {
                 topic.SubscriptionLock.EnterReadLock();
 
+                _trace.TraceVerbose("Scheduling topic '{0}'. Number of subscriptions: {1}", eventKey, topic.Subscriptions.Count);
```
davidfowl: This log is pretty useless.

src/Microsoft.AspNet.SignalR.Core/Messaging/ScaleoutMessageBus.cs
```diff
@@ -203,10 +207,25 @@ private void OnReceivedCore(int streamIndex, ulong id, ScaleoutMessage scaleoutM
             // Schedule after we're done
             foreach (var eventKey in keys)
             {
+                _trace.TraceVerbose("Scheduling event with key: '{0}'", eventKey);
```
davidfowl: Can we instead log all in a single line  and check if verbose is enabled?

src/Microsoft.AspNet.SignalR.Core/Messaging/MessageBus.cs
moozzyk: It tells me how many subscriptions subscribed to a given topic. Given that I am not exactly sure what I am looking for it might not be that useless. I would prefer to keep it.

moozzyk:
@davidfowl :up::date:

src/Microsoft.AspNet.SignalR.Core/Messaging/MessageBus.cs
davidfowl: It'll be super spammy and the number won't tell you much

moozzyk: Will remove and hope you are right :D

davidfowl:
👍

Issue #3895, opened 2017-03-22
Console application using async/await stops receiving after sending


markrendle:
I was working on a spike using a pair of console applications to send and receive messages, and hit a bit of odd behavior. In the subscriber application, once it had invoked a method on the Hub using an `await`, it stopped receiving new messages. Initially this meant it never received messages because it called a `Subscribe` method right away, but digging into it I found that it would receive quite happily _until_ it called a method.

When the method is invoked with a `.Wait()` instead of `await` everything is fine.

I'm not sure if this is to do with the way threads are handled in console applications (as opposed to WPF applications which have a `SynchronizationContext`), or is down to a problem in the C# SignalR client.

### Expected  behavior
Would expect to be able to `await hubProxy.Invoke("Foo");` and continue to receive messages.

### Actual behavior
Using `await hubProxy.Invoke("Foo");` breaks the connection somehow.

### Steps to reproduce

```c#
class Program
{
  static void Main()
  {
    Run().Wait();
  }

  static async Task Run()
  {
    Console.WriteLine("[Subscriber] press Enter to enter...");
    Console.ReadLine();

    var hubConnection = new HubConnection("http://localhost:39103/");
    var hubProxy = hubConnection.CreateHubProxy("MyHub");
    hubProxy.On<string>("OnMessage", msg =>
    {
      Console.WriteLine($"Received: {msg}");
    });

    await hubConnection.Start();

    // Receiving messages quite happily at this point

    Console.WriteLine("And press Enter again to break the application...");
    Console.ReadLine();

    await hubProxy.Invoke("Foo", "Bar");

    // NO MESSAGES FOR YOU!

    Console.WriteLine("And press Enter to Exit, because computers...");
    Console.ReadLine();
  }
}
```

markrendle:
I should point out, the reason I'm concerned about this even though it's just a spike is that the actual service I'm working on will be consumed from ASP.NET Web API services as well as WPF desktop applications.

moozzyk:
It's a very implicit deadlock. If you turn on tracing you will see that the messages are still being received by the client only your callback is not being invoked because the previous one has not finished yet: 
```
And press Enter to Exit, because computers...
16:44:27.6966075 - b8022975-3d42-4935-952d-7c43a392d049 - WS: OnMessage({"C":"d-59F20992-B,18|S,0|T,1","M":[{"H":"ChatHub","M":"broadcastMessage","A":["adfasf","sadf"]}]})
[...]

PR #3902, opened 2017-03-30
Fix for #3895

moozzyk:
Fixing an issue where user's code would run a continuation of `HubProxy.Invoke` on a SignalR thread because we were using TaskCompletionSource with default settings. If the user's code blocked it would completely block the receive queue and the user will no longer get any notifications even though the client processes them and add to the queue.  The fix is to dispatch the completion of `HubProxy.Invoke` to a different thread (note that we can't use `TaskCreationOptions.RunContinuationsAsynchronously` because some platforms we compile for do not support this).

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Infrastructure/DispatchingTaskCompletionSource.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Infrastructure/DispatchingTaskCompletionSource.cs
@@ -1,0 +1,72 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.SignalR.Infrastructure
+{
+    internal class DispatchingTaskCompletionSource<TResult>
+    {
+        private readonly TaskCompletionSource<TResult> _tcs = new TaskCompletionSource<TResult>();
+
+        public Task<TResult> Task
+        {
+            get { return _tcs.Task; }
+        }
+
+        public void SetCanceled()
+        {
+            TaskAsyncHelper.Dispatch(() => _tcs.SetCanceled());
+        }
+
+        public void SetException(Exception exception)
+        {
+            TaskAsyncHelper.Dispatch(() => _tcs.SetException(exception));
+        }
+
+        public void SetException(IEnumerable<Exception> exceptions)
+        {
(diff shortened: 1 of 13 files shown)


moozzyk:
@davidfowl - I am thinking about doing this everywhere in the code. What do you think? (I am actually wondering if this might be the root cause of random hangs in Groups.Add we have never been able to reproduce)

davidfowl:
Might be cleaner to make a type that wraps `TaskCompletionSource`. That's much easier to spot in code reviews.

`SignalRTaskCompletionSource`

moozzyk:
OK will try that.

moozzyk:
:up::date:

src/Microsoft.AspNet.SignalR.Core/TaskAsyncHelper.cs
```diff
@@ -1354,5 +1354,24 @@ private static class TaskCache<T>
         {
             public static Task<T> Empty = MakeTask<T>(default(T));
         }
+
+        [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode", Justification = "This is a shared file")]
+        public static void Dispatch(Action action)
```
davidfowl: A future version could avoid the closure in some cases by supporting a state parameter and `Action<T>`.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/DispatchingTaskCompletionSource.cs
```diff
@@ -0,0 +1,72 @@
...
+    internal class DispatchingTaskCompletionSource<TResult>
+    {
+        private readonly TaskCompletionSource<TResult> _tcs = new TaskCompletionSource<TResult>();
+
+        public Task<TResult> Task
+        {
+            get { return _tcs.Task; }
+        }
+
+        public void SetCanceled()
```
davidfowl: Do we use set* anywhere?

moozzyk: Not at the moment. I just tried to make it as similar to TaskCompletionSource as possible. We might need it when replacing TaskCompletionSources used on the server side. I will leave it for now and will replace TaskCompletionSources on the server side with this one and then remove methods that are not used/needed.

src/Microsoft.AspNet.SignalR.Core/TaskAsyncHelper.cs
moozzyk: OK. 
On a different note - the suppression should be removed since it is no longer correct/needed.

PR #3906, opened 2017-04-04
Replacing TaskCompletionSource with DispatchingTaskCompletionSource

moozzyk:
to prevent user's code to be run on SignalR threads

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Hubs/HubDispatcher.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Hubs/HubDispatcher.cs
@@ -192,7 +192,7 @@ namespace Microsoft.AspNet.SignalR.Hubs
         {
             // TODO: Make adding parameters here pluggable? IValueProvider? ;)
             HubInvocationProgress progress = GetProgressInstance(methodDescriptor, value => SendProgressUpdate(hub.Context.ConnectionId, tracker, value, hubRequest), Trace);
-            
+
             Task<object> piplineInvocation;
             try
             {
@@ -303,7 +303,7 @@ namespace Microsoft.AspNet.SignalR.Hubs
         [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "A faulted task is returned.")]
         internal static Task<object> Incoming(IHubIncomingInvokerContext context)
         {
-            var tcs = new TaskCompletionSource<object>();
+            var tcs = new DispatchingTaskCompletionSource<object>();
 
             try
             {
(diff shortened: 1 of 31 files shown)


src/Microsoft.AspNet.SignalR.Core/Hubs/HubDispatcher.cs
```diff
@@ -549,6 +549,55 @@ private Task ProcessResponse(StateChangeTracker tracker, object result, HubReque
             return Transport.Send(hubResult);
         }
 
+        // This method is being used to buld a callback with reflection
+        private static void ContinueWith<T>(Task<T> task, DispatchingTaskCompletionSource<object> tcs)
```
moozzyk: I removed this code in the previous checkin because it was unreferenced. Turned out that we use reflection and this is not dead code. Scary thing is that none of the tests run as part of build.cmd failed. I only noticed it because I ran functional tests I just fixed.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/DispatchingTaskCompletionSource.cs
```diff
@@ -46,9 +41,17 @@ public void TrySetException(Exception exception)
             TaskAsyncHelper.Dispatch(() => _tcs.TrySetException(exception));
         }
 
-        public void TrySetException(IEnumerable<Exception> exceptions)
+        public void SetUnwrappedException(Exception e)
```
davidfowl: Why isn't this dispatched?

src/Microsoft.AspNet.SignalR.Core/Hubs/HubDispatcher.cs
davidfowl: Ah yes, we do call this 😄

davidfowl:
Did you double check to make sure there's no places on the server where we were relying on the TCS *not* dispatching?

src/Microsoft.AspNet.SignalR.Core/Infrastructure/DispatchingTaskCompletionSource.cs
halter73: It wasn't changed in this PR, but TrySetUnwrappedException also doesn't dispatch. I think they both should.

src/Microsoft.AspNet.SignalR.Core/Infrastructure/DispatchingTaskCompletionSource.cs
```diff
@@ -26,11 +26,6 @@ public void SetException(Exception exception)
             TaskAsyncHelper.Dispatch(() => _tcs.SetException(exception));
```
halter73: Should DbOperation.ExecuteNonQueryAsync and SqlReceiver also be updated to use DispatchingTaskCompletionSource?

moozzyk:
:up::date:

I couldn't find any place where we rely on TCS not dispatching. One reason I wanted to be able to run functional tests was to possibly detect a case like this. (and added @halter73 to the review in case he remembers if we might have any occurrences)

PR #3952, opened 2017-06-17
Added FoundationHttpClient to UWP Client assembly.

### LINKED ISSUE #3951: Cannot use custom SSL certificate validation callback in UWP Client
Both [ServicePointManager.ServerCertificateValidationCallback](https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.servercertificatevalidationcallback(v=vs.110).aspx) and [HttpClientHandler.ServerCertificateCustomValidationCallback](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback) are unsupported on UWP platform. For custom SSL certificate validation UWP platform developers should use [Windows.Web.Http.HttpClient](https://docs.microsoft.com/en-us/uwp/api/windows.web.http.httpclient) with validation based on [HttpBaseProtocolFilter](https://docs.microsoft.com/en-us/uwp/api/windows.web.http.filters.httpbaseprotocolfilter).

### Expected  behavior
Developer being able to use self-signed certificates with invalid domain names for developing purposes and in production environments where DNS server in not available.

### Actual behavior
There is no support in built-in client for such functionality, thus developer must write client class from scratch.

### Steps to reproduce
Follow instruction [here](https://github.com/YakhontovYaroslav/FoundationHttpClientDemo)

YakhontovYaroslav:
- Added FoundationHttpClient to UWP Client assembly. This client is based on [Windows.Web.Http.HttpClient](https://docs.microsoft.com/en-us/uwp/api/windows.web.http.httpclient) and allowing developers to use things like custom SSL certificate validation and  usage of [ControlChannelTrigger](https://docs.microsoft.com/en-us/uwp/api/windows.networking.sockets.controlchanneltrigger) to build always reachable UWP clients.

This one fixes #3951


anurse:
We're moving to a model of having a single .NET Standard client that should support this custom callback.

Issue #3963, opened 2017-07-09
Bug with GetTypesSafe(Assembly a) in ReflectedHubDescriptorProvider


bklooste:
### Expected  behavior
No thrown exception only trace

### Actual behavior
 System.NotSupportedException thrown on a.Location

            try
            {
                return a.GetTypes();
            }
            catch (ReflectionTypeLoadException ex)
            {
                _trace.TraceWarning("Some of the classes from assembly \"{0}\" could Not be loaded when searching for Hubs. [{1}]\r\n" +
                                    "Original exception type: {2}\r\n" +
                                    "Original exception message: {3}\r\n",
                                    a.FullName,
                                    a.Location,
                                    ex.GetType().Name,
                                    ex.Message);
### Steps to reproduce
In a program that does some xml serialization call this after the serialization..

It is also likely that people who got this error for debugging for the same reason, 

Workaround .. In some cases you can call it early before the cache is built this is why most people dont see the error.

I have created a trivial pull https://github.com/SignalR/SignalR/pull/3964

moozzyk:
We should refine https://github.com/SignalR/SignalR/pull/3368.

moozzyk:
#3657

IGx89:
A production app of mine just ran into this as well, immediately after a daily app pool recycle. Recycling the app pool again resolved it.

Since this has apparently been a known issue since 2014 I'm not holding out hope it'll be fixed soon, however I'm curious if the .NET Core version of SignalR is affected or not by this issue? I would guess yes, due to how major of refactor it is, but if not we'd definitely want to make sure this issue is on their radar as well.

moozzyk:
I put it to the vNext milestone so that it is considered in the next servicing release.

liammycroft:
Will this need to be re-raised based on the techical bankruptcy decalred in https://blogs.msdn.microsoft.com/webdev/2018/09/17/the-future-of-asp-net-signalr/

anurse:
Normally, yes, it would need to be pinged and re-opened (as you've noticed we haven't done the closing yet)

[...]

PR #3964, opened 2017-07-09
Fix for issue  #3963 GetTypesSafe in ReflectedHubDescriptorProvider

bklooste:
Issue for exception handler throwing exception trying to trace.


moozzyk:
This fix is not correct - it swallows the exception so you don't even know that something happened. There is a PR to fix the same issue which does not swallow the exception: https://github.com/SignalR/SignalR/pull/3368.

bklooste:
Errors in tracing/ logging in a swallow exception handler **should** be swallowed ..it can be a whole bunch of random issues eg in this case its location I can have any 3rd party assembly that does something else eg NotImplemented on GetTypes why should this bring the whole of SignalIr down ?

PR #3971, opened 2017-07-19
removes typo

christopheralcock:


index.html
```diff
@@ -84,7 +84,7 @@ <h2 id="project_tagline">Incredibly simple real-time web for .NET</h2>
 
 <h2>Questions?</h2>
 
-<p>The SignalR team hangs out in the <a href="http://jabbr.net/#/rooms/signalr" target="_blank">signalr</a> room at on <a href="http://jabbr.net/">JabbR</a>.</p>
+<p>The SignalR team hangs out in the <a href="http://jabbr.net/#/rooms/signalr" target="_blank">signalr</a> room on <a href="http://jabbr.net/">JabbR</a>.</p>
```
moozzyk: I think this section should be replaced entirely. Jabbr is defunct

christopheralcock: very good point

PR #3982, opened 2017-07-31
Scaleout redis multi instance

aydar-musin:
Current implementation of Redis cluster is not optimized for scaling out (https://github.com/antirez/redis/issues/2672) . That is why I implemented  the version of RedisMessageBus (`MultiInstanceRedisMessageBus`) able to connect to several independent Redis instances. It subscribes for messages from all the instances in `ConnectToRedisAsync` method and it sends messages to the random instance in `Send` method. So load is distributed over the connected instances.


aydar-musin:
It doesn't work due to RedisMessage.Id is different for every redis instance. Thinking about how to solve it

PR #3983, opened 2017-08-04
Don't use deprecated in jquery v3 .bind/.unbind methods

PaulGiletich:
- also update jquery version from 1.6.2 to 1.7.2, where .on/.off is introduced

As for jquery v3, .bind/.unbind methods are deprecated in favour of .on/.off methods - https://jquery.com/upgrade-guide/3.0/#deprecated-bind-and-delegate . So when using signalR with jquery 3.x, console warnings are triggered. This PR fixes it.


moozzyk:
We cannot update dependency versions in a minor release - it's a breaking change.

thedutchess:
Any news on this please?

Issue #4020, opened 2017-10-04
Crank doesn't let you pass query strings


insulind:
### Expected  behavior
- To be able to load test a signalR endpoint that requires query string params passed to it
### Actual behavior
- There is currently no parameter to pass query strings to crank. If the url passed contains a question mark it crashes as there is an unhandled aggregate exception that contains the original exception that says, url can't contain a '?'
### Steps to reproduce
- Pass a url to crank with a query string such as http://mywebservice.com:8080/signalR?param=value and it doesnt work

I needed this functionality for some testing I was doing and implemented it myself. I would love to contribute to project, so if people are happy I would happily make an attempt at this. Let me know!

moozzyk:
We do accept contributions so feel free to send a PR.

insulind:
Cool, I'll get one in asap. Thanks!

insulind:
@moozzyk  I've got the dev branch and built it and I've ran the tests and some are failing. Is that to be expected? (I haven't made my changes fyi)

moozzyk:
You are likely missing some dependencies. Can you show the errors? (I think functional tests needed IIS manager and you may need to install some UWP tooling if it is not installed)

insulind:
```
Test Name:	Microsoft.AspNet.SignalR.Client.Tests.ConnectionFacts+Start.AwaitingOnFailedStartAndThenStoppingDoesntHang(hostType: Memory, transportType: LongPolling, messageBusType: Default)
Test FullName:	Microsoft.AspNet.SignalR.Client.Tests.ConnectionFacts+Start.AwaitingOnFailedStartAndThenStoppingDoesntHang (ae35e591e2a092a0e79470de8f84ec11011e89a3)
Test Source:	D:\SignalR\tests\Microsoft.AspNet.SignalR.FunctionalTests\Server\Connections\ConnectionFacts.cs : line 555
Test Outcome:	Failed
Test Duration:	0:00:21.06

Result StackTrace:	
at Microsoft.AspNet.SignalR.Client.Tests.ConnectionFacts.Start.<AwaitingOnFailedStartAndThenStoppingDoesntHang>d__13.MoveNext() in D:\SignalR\tests\Microsoft.AspNet.SignalR.FunctionalTests\Server\Connections\ConnectionFacts.cs:line 570
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #4023, opened 2017-10-06
Feature #4020

insulind:

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Crank/DefaultConnectionFactory.cs
+++ b/src/Microsoft.AspNet.SignalR.Crank/DefaultConnectionFactory.cs
@@ -14,7 +14,39 @@ namespace Microsoft.AspNet.SignalR.Crank
 	{
 		public Connection CreateConnection(string url)
 		{
-			return new Connection(url);
+			return new Connection(StripQueryString(url), GetQueryStringFromUrl(url));
 		}
-	}
+
+        /// <summary>
+        /// returns the url without the query and the query string as two seperate items
+        /// </summary>
+        /// <param name="url"></param>
+        /// <returns>Item1:The url without the query string
+        /// Item2:The query string if there was one or null</returns>
+        private static string GetQueryStringFromUrl(string url)
+        {
+            if (url == null)
+            {
+                throw new ArgumentNullException("url");
+            }
+
+            var firstIndex = url.IndexOf("?");
+            return firstIndex > -1 ? url.Substring(firstIndex + 1) : null;
+        }
+
+        /// <summary>
+        /// returns the url without the query string if one is present else it returns the original url
+        /// </summary>
+        /// <param name="url"></param>
+        /// <returns>The url without the query string or the original url if no query string was present</returns>
+        private static string StripQueryString(string url)
+        {
+            if (url == null)
+            {
+                throw new ArgumentNullException("url");
+            }
+            var firstIndex = url.IndexOf("?");
+            return firstIndex > -1 ? url.Substring(0, firstIndex) : url;
+        }
+    }
 }
--- a/src/Microsoft.AspNet.SignalR.Crank/CrankArguments.cs
+++ b/src/Microsoft.AspNet.SignalR.Crank/CrankArguments.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNet.SignalR.Crank
         [CommandLineParameter(Command = "Connections", Required = false, Default = 100000, Description = "(Connect phase) Number of connections to open. Default: 100000")]
         public int Connections { get; set; }
 
-        [CommandLineParameter(Command = "ConnectTimeout", Required = false, Default = 300, Description = "(Connect phase) Timeout in milliseconds. Default: 300 ms")]
+        [CommandLineParameter(Command = "ConnectTimeout", Required = false, Default = 300, Description = "(Connect phase) Timeout in seconds. Default: 300")]
         public int ConnectTimeout { get; set; }
 
         [CommandLineParameter(Command = "MinServerMBytes", Required = false, Default = 500, Description = "(Connect phase) Minimum server available MBytes to reach. Default: 500 MB")]
@@ -47,7 +47,7 @@ namespace Microsoft.AspNet.SignalR.Crank
         [CommandLineParameter(Command = "SendInterval", Required = false, Default = 500, Description = "(Send phase) Time in milliseconds between sends. Default: 500 ms")]
         public int SendInterval { get; set; }
 
-        [CommandLineParameter(Command = "SendTimeout", Required = false, Default = 300, Description = "(Send phase) Timeout in milliseconds. Default: 300 ms")]
+        [CommandLineParameter(Command = "SendTimeout", Required = false, Default = 300, Description = "(Send phase) Timeout in seconds. Default: 300")]
         public int SendTimeout { get; set; }
 
         [CommandLineParameter(Command = "ControllerUrl", Required = false, Description = "Url where one client will host a controller hub. Default: no controller (single client)")]


moozzyk:
Public API cannot be changed. You will need to revert changes to IConnection.cs

insulind:
will do

src/Microsoft.AspNet.SignalR.Crank/PerformanceCounters.cs
```diff
@@ -88,9 +88,9 @@ private static PerformanceCounter LoadCounter(string category, string name, stri
                 counter.NextSample();
                 return counter;
             }
-            catch (Exception)
+            catch (Exception x)
             {
-                Console.WriteLine("Unable to load counter '{0}\\{1}' on host '{2}'", category, name, host);
```
insulind: My performance counters were loading, which I think is a permissions issue, however i wasnt being told that. this will now give a bit more info

insulind:
It has occurred to me that I have also changed IConnectionFactory, is that a no go to?

moozzyk:
Yes. We should not public API especially that this would be shipped in a servicing release (I understand it is a tool but still this is the general rule we should try to respect unless there is a really good reason to not to). I also think passing the query string as a separate parameter does not buy you anything - it actually makes the usage of the tool worse. I think the idea here should be to split the url on the first `?` in url if one exists and take it from there.

insulind:
Yes thinking about I agree. To make it work for my situation this was easier, but I definitely agree this would be much better. I'll refactor this hopefully in the next few days abd make sure it's not affecting public APIs and minimal chance to behaviour. Thanks for the feedback

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -137,10 +137,11 @@ public class Connection : IConnection, IDisposable
         /// </summary>
         /// <param name="url">The url to connect to.</param>
         public Connection(string url)
-            : this(url, (string)null)
+            :this(url, GetQueryStringFromUrl(ref url))
```
moozzyk: I am not huge fan of this `ref` - Can you have two methods one that extracts the part before '?' and the other one that returns query string? This is not a perf sensitive code so it should be fine to scan the string twice.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -975,6 +976,26 @@ private static string CreateQueryString(IDictionary<string, string> queryString)
...
+        /// If a query string is not in the url it returns null and leaves the url untouched</returns>
+        private static string GetQueryStringFromUrl(ref string url)
+        {
+            if (url == null) throw new ArgumentNullException("url");
+            string queryString = null;
+            if (url.Contains("?"))
+            {
+                var urlSplitBits = url.Split('?');
+
+                queryString = queryString = urlSplitBits[1];
```
moozzyk: why do you need to set query string twice?

moozzyk: This would not work if there was another question mark in the query string - I think the best way is to find the index of the first '?' and if it is not -1 return a substring starting after this `?`

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -975,6 +976,26 @@ private static string CreateQueryString(IDictionary<string, string> queryString)
...
+        /// Gets the query from a url and removes the query string from the url passed
+        /// </summary>
+        /// <param name="url"></param>
+        /// <returns>the query string from the url passed and removes the query string from the url.
+        /// If a query string is not in the url it returns null and leaves the url untouched</returns>
+        private static string GetQueryStringFromUrl(ref string url)
+        {
+            if (url == null) throw new ArgumentNullException("url");
+            string queryString = null;
+            if (url.Contains("?"))
```
moozzyk: This probably should not be needed -  you could Split and check the number of items in the returned array. As commented below - I think Split may not work in some cases so I would recommend finding the index of first `?` and splitting on the index (if not -1)

src/Microsoft.AspNet.SignalR.Client/Connection.cs
insulind: not sure whats happened there, that wasn't intentional. I'll remove one of them

insulind:
Hi @moozzyk, in short I think I pushed this accidently so apologies there. I've been away for a few days and I've now made some changes that should hopefully be a bit better. Also turns out not only is the ref argument a bit of a code smell, it actually does work as expected the url argument has been captured already at that point so changes to it in that method called for the second argument doesn't make any difference, so I've learnt something there. I'll get what I've got pushed up, just running the rests atm

moozzyk:
@insulind  - it was the first time I saw `ref` used like that and was not sure myself if it really worked. This is why I asked to change it - so that I can understand the code :D

insulind:
@moozzyk should be good for a second review now, thanks again

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -137,10 +137,11 @@ public class Connection : IConnection, IDisposable
         /// </summary>
         /// <param name="url">The url to connect to.</param>
         public Connection(string url)
-            : this(url, (string)null)
+            :this(StripQueryString(url), GetQueryStringFromUrl(url))
```
moozzyk: One thing I did not notice before - the splitting should happen in https://github.com/SignalR/SignalR/blob/1fba14fa3437e24c204dfaf8a18db3fce8acad3c/src/Microsoft.AspNet.SignalR.Crank/DefaultConnectionFactory.cs#L17. Doing this in Connection.cs changes the current behavior which I don't want to avoid (I don't know how big/small it is but we should not change it)

Actuall now I see - before we would throw if url contained `?` 
https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Client/Connection.cs#L166 and after this change we wouldn't but `Connection` should not be changed for this class.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -975,6 +976,40 @@ private static string CreateQueryString(IDictionary<string, string> queryString)
...
+        /// <returns>Item1:The url without the query string
+        /// Item2:The query string if there was one or null</returns>
+        private static string GetQueryStringFromUrl(string url)
+        {
+            if (url == null) throw new ArgumentNullException("url");
+            string queryString = null;
+            var firstIndex = url.IndexOf("?");
+            if (firstIndex > -1)
+            {
+                queryString = url.Substring(firstIndex + 1);
```
moozzyk: just return instead of assigning - this is the pattern we use.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -975,6 +976,40 @@ private static string CreateQueryString(IDictionary<string, string> queryString)
...
 
+        /// <summary>
+        /// returns the url without the query and the query string as two seperate items
+        /// </summary>
+        /// <param name="url"></param>
+        /// <returns>Item1:The url without the query string
+        /// Item2:The query string if there was one or null</returns>
+        private static string GetQueryStringFromUrl(string url)
+        {
+            if (url == null) throw new ArgumentNullException("url");
```
moozzyk: nit: change to:
```
if (url == null)
{
  throw new ArgumentNullException("url");
}
```

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -975,6 +976,40 @@ private static string CreateQueryString(IDictionary<string, string> queryString)
...
+            string queryString = null;
```
moozzyk: this won't be needed once you return instead of assigning.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
moozzyk: or use ternary `return firstIndex > -1 ? queryString = url.Substring(firstIndex + 1) : null`

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -975,6 +976,40 @@ private static string CreateQueryString(IDictionary<string, string> queryString)
...
+        /// returns the url without the query string if one is present else it returns the original url
+        /// </summary>
+        /// <param name="url"></param>
+        /// <returns>The url without the query string or the original url if no query string was present</returns>
+        private static string StripQueryString(string url)
+        {
+            var firstIndex = url.IndexOf("?");
+            if (firstIndex > -1)
+            {
+                url = url.Substring(0, firstIndex);
```
moozzyk: Same as above - don't assign - just return

src/Microsoft.AspNet.SignalR.Client/Connection.cs
insulind: Yes that makes much more sense, I hadn't considered making the change there since we'd reverted the query string crank argument addition. I'll work on these changes tomorrow. Thanks

insulind:
@moozzyk I've done those changes. The splitting occurs in the DefaultConnectionFactory. Since that logic has been moved out of Connection the unit tests I had created are no longer relevant. I can't see where the tests for the crank project are located, are there any? If not should I create a new test project in the tests folder?

Thanks

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -137,10 +137,10 @@ public class Connection : IConnection, IDisposable
         /// </summary>
         /// <param name="url">The url to connect to.</param>
         public Connection(string url)
```
moozzyk: revert changes to this file

src/Microsoft.AspNet.SignalR.Crank/DefaultConnectionFactory.cs
```diff
@@ -14,7 +14,41 @@ class DefaultConnectionFactory : IConnectionFactory
...
+        private static string GetQueryStringFromUrl(string url)
+        {
+            if (url == null)
+            {
+                throw new ArgumentNullException("url");
+            }
+
+            var firstIndex = url.IndexOf("?");
+            return firstIndex > -1 ? url.Substring(firstIndex + 1) : null;
+
```
moozzyk: nit: remove empty line

src/Microsoft.AspNet.SignalR.Crank/DefaultConnectionFactory.cs
```diff
@@ -14,7 +14,41 @@ class DefaultConnectionFactory : IConnectionFactory
...
+        /// <returns>The url without the query string or the original url if no query string was present</returns>
+        private static string StripQueryString(string url)
+        {
+            if (url == null)
+            {
+                throw new ArgumentNullException("url");
+            }
+            var firstIndex = url.IndexOf("?");
+            return firstIndex > -1 ? url.Substring(0, firstIndex) : url;
+
```
moozzyk: nit: remove empty line

tests/Microsoft.AspNet.SignalR.Client.Tests/Client/ConnectionFacts.cs
```diff
@@ -439,5 +439,6 @@ public void OnReconnectedInvokesReconnectedOnHeartBeatMonitor()
 
             mockTransport.Verify(t => t.Dispose(), Times.Once());
         }
+
```
moozzyk: revert

moozzyk:
Thanks for the PR!

insulind:
Woo thanks! And thank you for the guidance it was appreciated :)

PR #4025, opened 2017-10-08
Fix: synchronous XMLHttpRequest.

### LINKED ISSUE #3624: chrome throws warning about Synchronous XMLHttpRequest
The issue is handled for Firefox 11+ but not for chrome.

Here is the code (line 589 of 2.2.0.js):
                        // Firefox 11+ doesn't allow sync XHR withCredentials: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#withCredentials
                        var isFirefox11OrGreater = signalR._.firefoxMajorVersion(window.navigator.userAgent) >= 11,
                            asyncAbort = !!connection.withCredentials && isFirefox11OrGreater;

Since I'm not using Firefox the var gets set to false, which then does a synchronous request causing a warning to be displayed in the console everytime I leave a webpage.

suggested fix: 
use the signalR async default here or create a new default for this exact situation

KyleScharnhorst:
- Was async for only Firefox, because firefox won't allow it to be
synchronous.
- No case where async shouldn't be an option here. Server would
eventually timeout after 30 seconds (default).
- Leaving firefox version check that handles different issue.
- asyncAbort set to true always.

fixes
#3624

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
+++ b/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.core.js
@@ -596,7 +596,7 @@
                     connection._.initHandler.start(transport, function () { // success
                         // Firefox 11+ doesn't allow sync XHR withCredentials: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#withCredentials
                         var isFirefox11OrGreater = signalR._.firefoxMajorVersion(window.navigator.userAgent) >= 11,
-                            asyncAbort = !!connection.withCredentials && isFirefox11OrGreater;
+                            asyncAbort = true;
 
                         connection.log("The start request succeeded. Transitioning to the connected state.");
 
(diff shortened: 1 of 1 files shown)


KyleScharnhorst:
Does this need a test?

anurse:
Thanks for the contribution! Sorry for the delay in responding, as we posted on [our blog](https://blogs.msdn.microsoft.com/webdev/2018/09/17/the-future-of-asp-net-signalr/) we let a bit of a backlog build up here, but we're cleaning that up now.

I don't think a test is necessary here since our existing functional tests should give us enough coverage h ere. If you can rebase your change on top of our latest `master`, I can get the tests to run on this PR and we can merge it. If you're not able or interested in following up, no worries, it's been almost a year 😰 . I can move the change to a new PR myself and get the tests running on it.

anurse:
Hey, as I mentioned above, if you're not able to follow up on this I'll just close this and open a new PR that's properly rebased. I should be able to maintain your author information though so you still get credit!

Thanks for your contribution! Feel free to comment further over on the updated PR (link TBD) if you have future comments!

Issue #4036, opened 2017-11-07
Intermittent Error: Cannot access a disposed object


MisterAcoustic:
### Description
Hello,

Our application undergoes automated testing after each commit/build. Once every two or three hundred tests, the testing will fail due to an exception in SignalR.

Apologies that I cannot adhere to the guidelines very closely due to the nature of the error. I do not know know how to reproduce the problem on demand. We are using version 2.2.2 of SignalR, testing on Windows Server 2016, IIS version 10.0.14393.0.

Here is a typical stack trace:

```
Cannot access a disposed object. Object name: \u0027SERVERNAME\u0027. 
at Microsoft.AspNet.SignalR.Messaging.ScaleoutStream.Send(Func2 send, Object state) 
at Microsoft.AspNet.SignalR.Infrastructure.Connection.Send(ConnectionMessage message) 
at Microsoft.AspNet.SignalR.Hubs.HubDispatcher.Outgoing(IHubOutgoingInvokerContext context) 
at Microsoft.AspNet.SignalR.Hubs.HubPipelineModule.\u003c\u003ec__DisplayClass6_0.\u003cBuildOutgoing\u003eb__0(IHubOutgoingInvokerContext context) 
at Microsoft.AspNet.SignalR.Hubs.SignalProxy.TryInvokeMember(InvokeMemberBinder binder, Object[] args, Object\u0026 result) at CallSite.Target(Closure , CallSite , Object , String ) 
at E2SSPro.Web.SignalR.DBUpdateHub.SendMessage(String msg, String signalToken) in D:\\TeamCity\\BuildAgent3\\work\\b28bc0f3a005363e\\E2SSPro.Web\\SignalR\\DBUpdateHub.cs:line 25 
at E2SSPro.Web.Controllers.DBUpdateController.AddProgressMsg(String message) in D:\\TeamCity\\BuildAgent3\\work\\b28bc0f3a005363e\\E2SSPro.Web\\Controllers\\DBUpdateController.cs:line 103 
at E2SSPro.Web.Controllers.DBUpdateController.UpdateStatusDelegate(String message, Boolean messageBox, Boolean logFile) in D:\\TeamCity\\BuildAgent3\\work\\b28bc0f3a005363e\\E2SSPro.Web\\Controllers\\DBUpdateController.cs:line 113 ...
```

###Background
[...]

PR #4059, opened 2018-01-04
Simplify Redis Connection logic and remove unnecessary state check #4036

anurse:
RedisMessageBus has some unncessary state management that causes problems if a Reconnect occurs in the middle of the `ConnectWithRetry` flow. This can happen (reported in #4036) and when it does, the connection becomes broken. This change should allow the connection to continue. There are no new locks, so I'm no longer concerned about deadlocks, some fairly simple smoke testing should be sufficient to give me confidence it's working. I'm going to merge fairly soon so that I can get a build out of the CI and get it to some of the customers who are seeing this issue.

FYI @misteracoustic @sreead as you've been on the Issue thread.

FYI @muratg 

Related: #4036

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
+++ b/src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
@@ -4,9 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
-using System.Globalization;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNet.SignalR.Messaging;
@@ -28,9 +25,18 @@ namespace Microsoft.AspNet.SignalR.Redis
 
         private IRedisConnection _connection;
         private string _connectionString;
-        private int _state;
         private readonly object _callbackLock = new object();
-        private readonly SemaphoreSlim _redisConnectionEventLock = new SemaphoreSlim(1, 1);
+
+        private int _state = State.Initial;
+
+        internal static class State
+        {
+            public const int Initial = 0;
+            public const int Connecting = 1;
+            public const int Connected = 2;
+            public const int Disposing = 3;
+            public const int Disposed = 4;
+        }
 
         public RedisMessageBus(IDependencyResolver resolver, RedisScaleoutConfiguration configuration, IRedisConnection connection)
             : this(resolver, configuration, connection, true)
@@ -47,6 +53,9 @@ namespace Microsoft.AspNet.SignalR.Redis
             }
 
             _connection = connection;
+            _connection.ConnectionFailed += OnConnectionFailed;
+            _connection.ConnectionRestored += OnConnectionRestored;
+            _connection.ErrorMessage += OnConnectionError;
 
             _connectionString = configuration.ConnectionString;
             _db = configuration.Database;
@@ -69,9 +78,6 @@ namespace Microsoft.AspNet.SignalR.Redis
 
         public TimeSpan ReconnectDelay { get; set; }
 
-        // For testing purposes only
-        internal int ConnectionState { get { return _state; } }
-
         public virtual void OpenStream(int streamIndex)
         {
             Open(streamIndex);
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
```diff
@@ -200,50 +204,70 @@ private void OnConnectionError(Exception ex)
...
+                        }
+
+                        break;
                     }
-                    else
+                    catch (Exception ex)
                     {
-                        Debug.Assert(oldState == State.Disposing, "unexpected state");
-                        _trace.TraceError("Unexpected state.");
+                        _trace.TraceError("Error connecting to Redis - " + ex.GetBaseException());
```
muratg: :+1:

src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs
```diff
@@ -200,50 +204,70 @@ private void OnConnectionError(Exception ex)
             }
             finally
             {
+                _trace.TraceVerbose($"{nameof(ConnectWithRetry)}: Releasing Redis Connection Event Lock");
```
moozzyk: Should be `nameof(OnConnectionRestored)`

anurse:
> I guess the thing I don't understand is how we got the Connection restored events without getting the OnConnectionFailed first

I'm assuming it's a further race condition between `ConnectAsync` and the binding of the events in `ConnectToRedisAsync`. We bind the events AFTER `ConnectAsync` which means that if the connection goes down immediately after `ConnectAsync`, it's possible the `ConnectionFailed` handler isn't bound. The logs in the issue indicate the reconnect is occurring immediately after connection, so it seems feasible.

> If this can happen, maybe the fix is to introduce a new Connecting state during which OnConnectionRestored events are ignored

Is it safe to ignore the OnConnectionRestored event or does it still need to be processed? It looks like it would probably be safe to drop it, as the only thing the event does is send the latest message ID to Redis (which is unnecessary if we're in the middle of the initial connection) and Open the stream (which again, `ConnectWithRetry` will do when it finishes). I'll iterate on that.

anurse:
Looking at the code, I think it's actually OK to allow all these events to fire without trying to serialize or synchronize them.

All the `OnConnectionFailed` event does is log that the connection was dropped and update the internal state value, but that value seems to be unnecessary (as I'll describe below).

The `OnConnectionRestored` event ensures that Redis has the latest message ID according to this server (updating the value only if the Redis ID value is lower than that node's value) and then signals back to the outer scale-out infrastructure that the stream is "up" again (but that's only actually necessary when a previous Send had actually failed). It also updates the state value, but again, that seems to be unnecessary.

So, running `OnConnectionFailed` at any time is generally safe, since other than updating the internal state value, it does nothing but logging. If the connection fails in the middle of our connection flow (i.e. after it has actually connected to Redis, but before we complete our ConnectWithRetry loop), it's harmless. It's a failure in `Send` that **actually** triggers the scale-out infrastructure to consider the stream down.

Running `OnConnectionRestored` at any time is also safe because it will perform a harmless atomic update to Redis in order to ensure that the message ID is up-to-date and then re-open the stream. This is no different to what would happen if the connection failed and was restored all in between two calls to `Send`. The `Open` method already has code to no-op if the stream is already open, so this is safe.

Given all that, I think the right fix here is probably to drop the locking and state flags entirely. Instead, we can just use a `_disposed` flag to indicate when `Dispose` has been called. We can also drop the `Shutdown` call [here](https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Redis/RedisMessageBus.cs#L229) because we no longer care about the state values, other than perhaps `_disposed`, which we could have a check for in that method to ensure that if the connection is disposed mid-connect then we log that and shutdown properly (though again, that's not really necessary since the call to `Dispose` that would have set `_disposed` will already have closed the connection)

I feel fairly confident this is the right behavior, but I'm not seeing much in the way of automation for Redis functional testing which means it's going to be a pain to try and verify this.

moozzyk:
Yeah, the ConnectionMultiplexer does not allow subscribing to events before the connection has already been opened. This pattern was always a bit concerning to me (btw. as you remember this was also the reason why we dropped the original idea of having a static `Connection.ConnectAsync` method that returned a connected connection - the user could have missed notifications that arrived between opening the connection and subscribing to events). 
I agree with your analysis. Keeping the state does not seem to be necessary and getting rid of will simplify the code. 
With regards to testing the new functionality and prevent regressions - maybe it makes sense to leave the old class in the next release and hide it behind a configuration flag so that people can fall back to it if it turns out that the new class have issues.

anurse:
> maybe it makes sense to leave the old class in the next release and hide it behind a configuration flag

@moozzyk Ye olde feature gate :). I like it.

anurse:
I'm not actually too worried about dropping this behavior having looked over it quite a bit. We'll have some folks test out the bits and see what happens, it's also super easy to restore the old code or insert a quirk as you indicated. I'll do some basic verification to make sure the main scenarios aren't broken though.

anurse:
:up: :date: No longer WIP. I'm working a little to do some final verification, but I expect to merge this shortly so I can get a build out of the CI to send to some folks who are seeing the issue.

anurse:
Did some verification of Redis scenarios, still working with this change. The build is really sensitive, so merging this to get a CI build and see if I can iterate with that.

Issue #4077, opened 2018-02-26
[Question] How can I map only specific hubs in SignalR


zorgoz:
I have two hubs in my application. For the one I need NTLM authentication, and IP based filtering for the other. For this to work I need to set HttpListener.AuthenticationSchemes to AuthenticationSchemes.Ntlm for the first, while it remains Anonymous for the other. This means that I can not have both hubs running on the same listener connection. I have my OwinMiddleware to handle the authentication, and I am trying to set up two different endpoints for these two hubs (two different ports). My problem is that when I don't know how to tell SignalR to map only a specific hub to specific IAppBuilder, and not all hubs that it can find. 

This question is unanswered on [StackOwerflow](https://stackoverflow.com/questions/46726692/how-can-i-map-only-specific-hubs-in-signalr) for half a year.

anurse:
You should be able to support both NTLM and Anonymous authentication in HttpListener by simply ORing the two values together (it's a Flags enum not a standard enum): https://docs.microsoft.com/en-us/dotnet/api/system.net.authenticationschemes?view=netframework-4.7.2

For example:

```csharp
httpListener.AuthenticationSchemes = AuthenticationSchemes.Ntlm | AuthenticationSchemes.Anonymous
```

If that doesn't work, can you provide a sample app that reproduces the issue so I can investigate futher?

anurse:
Ah, I see what you're looking for. That would be a pretty major change and we don't have plans to do that in the upcoming 2.4.0 wave of releases. In fact, it's these limitations that are precisely why ASP.NET Core SignalR allows this kind of mapping of different hubs to different endpoints.

Anyway, I'm going to close this. We don't have plans to make this significant a change to the architecture of ASP.NET SignalR at this point.

PR #4094, opened 2018-04-04
Increasing level of extensibility

zorgoz:
Aside some minor maintainability related changes, I have made `ReflectedHubDescriptorProvider.BuildHubsCache` method virtual. This way one can extend the default provider in an easy way, without the need to recreate all its functionality. Because `HubTypeExtensions` were `internal`, one would have had to recreate that too. This is why I have made the latter `public`.

#4077 These changes provide a solution for this issue by supporting the creation of such providers:
```
public class FilteredHubDescriptorProvider : ReflectedHubDescriptorProvider
{
	private readonly Predicate<HubDescriptor> _filter;

	public FilteredHubDescriptorProvider(IDependencyResolver resolver, Predicate<HubDescriptor> filter): base(resolver)
	{
		_filter = filter;
	}

	protected override IDictionary<string, HubDescriptor> BuildHubsCache()
	{
		return base.BuildHubsCache().Where(d => _filter(d.Value)).ToDictionary(x => x.Key, x=> x.Value);
	}
}
```


src/Microsoft.AspNet.SignalR.Core/Hubs/Extensions/HubTypeExtensions.cs
```diff
@@ -5,9 +5,9 @@
 
 namespace Microsoft.AspNet.SignalR.Hubs
 {
-    internal static class HubTypeExtensions
```
anurse: We can't really make these public as extension methods on `Type`. Unless there's an extremely good reason, we try to avoid attaching public extension methods on public BCL types like this because they appear for everyone using the type if they happen to have our namespace imported. (If we were writing these helpers now, we wouldn't have made them extension methods at all, even though they are `internal` 😉).

Can you either leave these as `internal` (and just copy the code for your own extensions) or convert them to standard static methods (which would require changing all the call-sites unfortunately).

src/Microsoft.AspNet.SignalR.Core/Hubs/Lookup/ReflectedHubDescriptorProvider.cs
```diff
@@ -11,133 +11,132 @@
...
-                                    a.FullName,
-                                    a.Location,
-                                    ex.GetType().Name,
-                                    ex.Message);
-
-                return Enumerable.Empty<Type>();
-            }
-        }
-    }
+	public class ReflectedHubDescriptorProvider : IHubDescriptorProvider
```
anurse: For future reference, please avoid major reformatting like this in your contributions. We have to reformat it back to our standard code style in order to merge it :). Don't worry about it for this change, I can review the actual functional change just fine and resolve the formatting when I merge, but just for the future :).

anurse:
Also, could you [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) your change on top of the latest copy of the `master` branch and then change the base branch of the PR in GitHub (after pushing your changes). To change the base branch of the PR, just click "Edit" in the top-right and change the branch from the drop down:

image

src/Microsoft.AspNet.SignalR.Core/Hubs/Extensions/HubTypeExtensions.cs
zorgoz: Git your point. It is not that important after all. I have discarded this change during rebase.

src/Microsoft.AspNet.SignalR.Core/Hubs/Lookup/ReflectedHubDescriptorProvider.cs
zorgoz: I have noticed that only after pushed the changes. I have reworked this during rebase.

zorgoz:
Well, now CI build fails. But looks not to have anything to do with my modifications. `samples\Microsoft.AspNet.SignalR.ServiceSample\Startup.cs` is using some `MapAzureSignalR` method, but it is nowhere defined in the original repository.

https://github.com/SignalR/SignalR/search?q=MapAzureSignalR

anurse:
Hrm strange. I'll take a look at it and resolve it. You're correct, it's not part of your change, it's something from the more recent builds.

anurse:
Looks like the Azure SignalR SDK team cleared out their nightly feed, but fortunately they've also published a preview of their package. Once https://github.com/SignalR/SignalR/pull/4218 is merged, we can rebase this on master and the CI should be fixed.

There are also some flaky functional tests (which I've been slogging through de-flaking) that may fail, so if the CI goes red, it may not mean we can't merge it 😄

anurse:
It should be possible to rebase this on `master` now and get the CI build running properly. Are you able to do that @zorgoz ?

Issue #4048, opened 2017-12-01
SignalR Subscriber client(.NET/Java Script) connected but not received message after few minutes  


msinha53:
Getting this type of issue on some machines.
Issue Details:
SignalR server deployed on Windows Server 2012. 
When client subscriber the topic and wait for data & Publisher client published data then subscriber received data.
Subscriber idle some time(5-10 minutes).
Again Publisher Client start to published data but didn't received at subscriber client side.    
Subscriber client already connected but unable to received data.
I have verified SignalR trace on client side and server side but not any issue detected.
If subscriber client rejoin the topic then able to received data.
I think some how topic released from groups but we are unable to find out the root cause.
Already verified IIS Logs & Event viewer logs but didn't found any issue.

Please help us..

Pokoto:
I am experiencing the same issue with version 2.2.2

- I have been running 2.2.1 smoothly for more than a year on multiple Windows Server 2012 machines
- simply upgraded the client and server to 2.2.2  - default settings with no tweaks.
- using the Redis backplane (RedisLabs redis_version 2.8.12)
- randomly, after 5-25 minutes idle js clients stop receiving messages BUT a) can still send messages over the connection, b) after sending them the connection is aborted and a reconnect triggered, eventually the client starts to receive messages again
- tried with multiple browser tabs connected to different servers BUT subscribed to the same topic, they stop receiving messages at the same time
- the issue affects WebSockets but haven't tried with other protocols yet
- can't repro on the dev machine, is working fine (W10/VS2015/IISEXPRESS)

had to rollback to 2.2.1 as it's failing silently on both client and server...

Any Idea? :)

davidfowl:
Turn client side logging on. That should provide more details.

msinha53:
I have already enable SignalR trace log on client side and server side but no any issue detected. 
For connection testing, i have created KeepAlive function under Hub and client called KeepAlive on every 1 minute interval time and connection is working fine.
Publisher is publishing message on 5 minute interval time but subscriber unable to received.
If i have started new instance of Subscriber and join the topic then received 1 time published message but didn't received next message.
I think some how topic expired/removed from Group.
Any idea Group have any expiration time???

MarcoCestari:
Same issue described by @Pokoto here.
Seems to occurs only with remote connections.

VladimirMakaev:
[...]

PR #4120, opened 2018-05-24
fix issue with ScaleoutStore minMappingId being incorrect

anurse:
The `TryAddImpl` method on ScaleoutStore doesn't properly update `_minMappingId`. This is mostly benign except for the one case where `_minMappingId` is used: to detect that a cursor is so far behind that they should just be given all messages when requesting new messages.

Instead of being set to a "Mapping ID", the `_minMappingId` value gets set to a local message ID whenever a fragment is overwritten. This means the value gets corrupted. It was hard to detect this in our tests because they use the same value for both Mapping ID and Local ID, so I changed the tests to have the Mapping IDs start much higher (which is usually the case when deep into a long-running application).

This issue can cause a long-idle connection to end up being frozen out and seeing no new messages because their cursor value (which might be very high) is never lower than the `_minMappingId` (because local IDs tend to be much smaller).

Fixes: https://github.com/SignalR/SignalR/issues/4121

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Messaging/ScaleoutSubscription.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Messaging/ScaleoutSubscription.cs
@@ -100,9 +100,6 @@ namespace Microsoft.AspNet.SignalR.Messaging
                 ScaleoutMapping mapping = enumerator.Current.Item1;
                 int streamIndex = enumerator.Current.Item2;
 
-                _trace.TraceVerbose("Extracting message for scaleout mapping: {0}, stream index {1}, [{2}]",
-                    mapping.Id, streamIndex, Identity);
-
                 ulong? nextCursor = nextCursors[streamIndex];
 
                 // Only keep going with this stream if the cursor we're looking at is bigger than
@@ -113,8 +110,6 @@ namespace Microsoft.AspNet.SignalR.Messaging
 
                     // Update the cursor id
                     nextCursors[streamIndex] = mappingId;
-                    _trace.TraceVerbose("Updated cursor for mapping id {0} stream idx {1} to {2} [{3}]",
-                        mapping.Id, streamIndex, mappingId, Identity);
                 }
             }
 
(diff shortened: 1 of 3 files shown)


anurse:
FYI, some tests are failing now with the new cursor values. I think I just derped them up, but I'll review all that.

anurse:
Tests updated and fixed, I had indeed just derped them up by only half-updating them to use the new `MappingIdBase` value.

src/Microsoft.AspNet.SignalR.Core/Messaging/ScaleoutSubscription.cs
```diff
@@ -100,9 +100,6 @@ protected override void PerformWork(IList<ArraySegment<Message>> items, out int
                 ScaleoutMapping mapping = enumerator.Current.Item1;
                 int streamIndex = enumerator.Current.Item2;
 
-                _trace.TraceVerbose("Extracting message for scaleout mapping: {0}, stream index {1}, [{2}]",
```
anurse: Extra noisy traces I had added in an earlier commit for debugging that need to be removed.

anurse:
Ping @halter73 @davidfowl - If I can get approval, I can merge this today and have a build ready Tuesday morning ;).

src/Microsoft.AspNet.SignalR.Core/Messaging/ScaleoutSubscription.cs
halter73: Are all the other traces you added quiet enough to keep?

anurse: All the other traces were removed in earlier commits ;)

anurse: But I'll do a quick pass to sanity check that.

MarcoCestari:
maybe this was the cause of #4048?

PR #4186, opened 2018-08-29
[SignalR Service Client]Add redirect support for js client

vicancy:
1. Add redirect support with `{AccessToken: ..., Url:...}` response
2. Add bearer authorization support, for ws and sse, token is added in the query string, for longpolling, header is used

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
@@ -20,6 +20,7 @@
         stoppedWhileLoading: "The connection was stopped during page load.",
         stoppedWhileNegotiating: "The connection was stopped during the negotiate request.",
         errorParsingNegotiateResponse: "Error parsing negotiate response.",
+        errorRedirectionExceedsLimit: "Negotiate redirection limit exceeded.",
         errorDuringStartRequest: "Error during start request. Stopping the connection.",
         stoppedDuringStartRequest: "The connection was stopped during the start request.",
         errorParsingStartResponse: "Error parsing start response: '{0}'. Stopping the connection.",
@@ -408,7 +409,7 @@
 
         state: signalR.connectionState.disconnected,
 
-        clientProtocol: "1.5",
+        clientProtocol: "2.0",
 
         reconnectDelay: 2000,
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -689,69 +695,112 @@
                         return;
                     }
 
-                    keepAliveData = connection._.keepAliveData;
-                    connection.appRelativeUrl = res.Url;
-                    connection.id = res.ConnectionId;
-                    connection.token = res.ConnectionToken;
-                    connection.webSocketServerUrl = res.WebSocketServerUrl;
+                    // redirect case
+                    if (res.AccessToken && res.Url) {
```
anurse: We'll need to set a new `ProtocolVersion` value as part of this. If an older client receives this message, it will either try to connect directly to the app (ignoring the `Url` field) or produce an unhelpful error message.

My suggestion would be to change this to `if(res.ProtocolVersion === "2.0")` and ensure the SDK adds that value to the response.

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -689,69 +695,112 @@
...
+                        // initiate the actual connection to res.Url with res.AccessToken
+                        connection.accessToken = res.AccessToken;
+                        setConnectionUrl(connection, res.Url);
```
anurse: This pattern will only support one cycle of redirection, whereas ASP.NET Core SignalR supports up to 100 levels of redirection. However, that's because we tried to view the ASP.NET Core feature as "redirection" rather than just "SignalR Service Support". For ASP.NET SignalR, I'm much less worried about the general-purpose redirection feature, so I think this is OK. Thoughts @davidfowl ?

anurse: Also, can we log here?

```
connection.log("Received redirect to: " + res.Url);
```

davidfowl: Practically there's only ever one but it would be nice to be consistent.

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
vicancy: https://github.com/SignalR/SignalR/blob/master/src/Microsoft.AspNet.SignalR.Core/Infrastructure/ProtocolResolver.cs should also be updated to support higher version

vicancy: change to use `RedirectUrl` to distinguish with already in-use `Url`

anurse:
Yep, I'll take a look. I doubt it's related. There are a few flaky tests I'm still working on fixing.

anurse:
I squashed and merged this manually but GitHub didn't follow that, so I'm closing this.

PR #4193, opened 2018-09-06
Add JS Tests and compatibility logic to allow new client to connect to old servers

anurse:
Adds a JS functional test for #4186 . It works great!

Also, we have to be a little more clever with the protocol version. Here are the scenarios we want and how we want them to work:

1. New Client (2.0) -> Old Server (1.4): The client will request 2.0 but the server will respond 1.4. We need the client to accept this result, since we are compatible with 1.4
2. Old Client (1.4) -> New Server (2.0) **not using** a redirect: The client will request 1.4 and the server will respond with 1.4.
3. Old Client (1.4) -> New Server (2.0) **using** a redirect: The server should respond with the 2.0 protocol version which will automatically cause the client to report this error: "You are using a version of the client that isn't compatible with the server. Client version 1.4, server version 2.0."

This PR should get all three to work properly and has tests for 2 and 3. Testing 1 is much harder in this repo.

FYI @vicancy


src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -714,15 +719,16 @@
...
                             protocolError = signalR._.error(signalR._.format(resources.protocolIncompatible, connection.clientProtocol, res.ProtocolVersion));
                             $(connection).triggerHandler(events.onError, [protocolError]);
                             deferred.reject(protocolError);
 
                             return;
                         }
 
-                        if (res.RedirectUrl) {
+                        // Check for a redirect response (which must have a ProtocolVersion of 2.0)
+                        if (res.ProtocolVersion == "2.0" && res.RedirectUrl) {
```
davidfowl: ===

test/Microsoft.AspNet.SignalR.Client.JS.Tests/Startup.cs
```diff
@@ -44,11 +45,54 @@ public void Configuration(IAppBuilder app)
...
                 EnableDefaultFiles = true,
                 FileSystem = new PhysicalFileSystem(Path.Combine(Directory.GetCurrentDirectory(), "wwwroot"))
             });
         }
+
+        private static Func<IOwinContext, Func<Task>, Task> CreateRedirector(string sourcePath, string targetPath, string protocolVersion = null)
+        {
+            return (context, next) =>
+            {
+                if(context.Request.Path.StartsWithSegments(new PathString(sourcePath)))
```
davidfowl: nit: spacing

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -411,6 +411,11 @@
 
         clientProtocol: "2.0",
 
+        // We want to support older servers since the 2.0 change is to support redirection results, which isn't
+        // really breaking in the protocol. So if a user updates their client to 2.0 protocol version there's
+        // no reason they can't still connect to a 1.4 server.
+        supportedProtocols: [ "1.4", "2.0" ],
```
vicancy: isn't it 1.5 for js?
https://blog.3d-logic.com/2015/03/29/signalr-on-the-wire-an-informal-description-of-the-signalr-protocol/
> clientProtocol – the version of protocol used by the client. The most recent version is 1.5 however it is only used by the JavaScript client since the change that mandated bumping the version of the protocol to 1.5 is only relevant for this client. The .NET and C++ clients currently use version 1.4. Note that the server is designed to support down-level clients (i.e. clients using previous versions of the protocol) and the current (2.2.0) version supports protocol versions from 1.2 to 1.5

vicancy: And why we only support "1.4" only instead of "1.2 to 1.5" for the *1. New Client -> Old Server* scenario?

davidfowl: The sever sends the client version back so 1.2 would never come back from this client.

vicancy: But we want to support the case when New Client(2.0) connects to Old Server(1.4 or others?), which can be 1.2?

davidfowl: Not generally not, we don't support new clients connecting to old servers when there's a protocol change. Notice the client's minimum version isn't changing to 2.0 here. Newer clients can support both the redirect based negotiation urls (2.0) and 1.4 endpoints.

anurse: Yep, this should be 1.5. Good catch, I'll update that.

As @davidfowl said, we normally do not permit a new client to connect to an old server (in the case of a protocol change). This is an exception because the protocol change is a very minor negotiation change and it's trivial to allow this in this case.

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -714,15 +719,16 @@
                             return;
                         }
 
-                        if (!res.ProtocolVersion || res.ProtocolVersion !== connection.clientProtocol) {
+                        if (!res.ProtocolVersion || (connection.supportedProtocols.indexOf(res.ProtocolVersion) === -1)) {
                             protocolError = signalR._.error(signalR._.format(resources.protocolIncompatible, connection.clientProtocol, res.ProtocolVersion));
                             $(connection).triggerHandler(events.onError, [protocolError]);
```
vicancy: Is it that there is no console log or error when this error happens? It fails silently when I test locally. Am I missing any config?
image

anurse: Yeah, we should put a log in, good catch :). I'll add that in before merging.

anurse: Actually there are already logs, but `connection.log` writes to `console.debug` which means it only shows up if "Verbose" messages are enabled:

image

davidfowl:
Why is the CI failing?

anurse:
You've got a mouse, and access to VSTS ;P. Still finding and bashing flaky tests.

anurse:
Merging this now to unblock other work.

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
vicancy: got it! Thanks for the explanation

PR #4194, opened 2018-09-07
Revive the .NET Functional Tests

anurse:
I dropped a lot of the absurd extra coverage that took the runtime up into the 30 minute range. Also, I think I got it to work with parallel test execution (may have some flaky tests to shake out in the next few builds).

Total execution time is down to about 2 mins now!


test/Microsoft.AspNet.SignalR.FunctionalTests/Client/LongPollingFacts.cs
```diff
@@ -103,15 +105,20 @@ public void LongPollingDoesNotPollAfterTransportIsBeingStoppedMidRequest()
...
 
-            Assert.True(
-                longPollingTransport.Start(mockConnection.Object, string.Empty, disconnectCts.Token)
-                    .Wait(TimeSpan.FromSeconds(15)));
+            try
+            {
+                await longPollingTransport.Start(mockConnection.Object, string.Empty, disconnectCts.Token)
+                    .OrTimeout(TimeSpan.FromSeconds(15));
+            }
+            catch (StartException stex) when (stex.InnerException is OperationCanceledException)
```
anurse: Paging @davidfowl . It seems like this test sometimes throws up an OperationCanceledException *because* of the `disconnectCts.Cancel()` line above, in the test. It only throws that sometimes though, and it seems to be expected that the Start would fail since we're stopping it mid-request.

test/Microsoft.AspNet.SignalR.FunctionalTests/Server/Hubs/HubProgressFacts.cs
```diff
@@ -20,30 +18,57 @@ public class HubProgressFacts : HostedTest
...
             using (var host = CreateHost(hostType, transportType))
             {
                 host.Initialize(messageBusType: messageBusType);
 
-                HubConnection hubConnection = CreateHubConnection(host);
-                IHubProxy proxy = hubConnection.CreateHubProxy("progress");
-                var progressUpdates = new List<int>();
+                var hubConnection = CreateHubConnection(host);
+                var proxy = hubConnection.CreateHubProxy("progress");
+                var progressUpdates = Channel.CreateUnbounded<int>();
```
anurse: Paging @davidfowl. This test was funky and flaky before. Now it's snazzy and stable. Instead of waiting some arbitrary interval and expected the progress data to be written, we wait *until* the data has been written (with some timeout value)

test/Microsoft.AspNet.SignalR.FunctionalTests/Client/WebSocketFacts.cs
```diff
@@ -33,7 +33,7 @@ public async Task ClientCanReceiveMessagesOver64KBViaWebSockets(HostType hostTyp
 
                     await connection.Start(host.Transport);
 
-                    var result = await hub.Invoke<string>("ReturnLargePayload").OrTimeout();
+                    var result = await hub.Invoke<string>("ReturnLargePayload").OrTimeout(TimeSpan.FromSeconds(30));
```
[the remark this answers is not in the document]
anurse: When running in parallel, it can actually take longer than the default 5 second timeout for this to execute because the CPU is busy with a bunch of other tests. It's still faster overall to run in parallel though.

test/Microsoft.AspNet.SignalR.FunctionalTests/Server/Hubs/HubProgressFacts.cs
```diff
@@ -20,30 +18,57 @@ public class HubProgressFacts : HostedTest
         //[InlineData(HostType.IISExpress, TransportType.ServerSentEvents, MessageBusType.Default, Skip = "Disabled IIS Express tests because they fail to initialize")]
         //[InlineData(HostType.IISExpress, TransportType.LongPolling, MessageBusType.Default, Skip = "Disabled IIS Express tests because they fail to initialize")]
         [InlineData(HostType.HttpListener, TransportType.Websockets, MessageBusType.Default)]
-        [InlineData(HostType.HttpListener, TransportType.ServerSentEvents, MessageBusType.Default)]
-        [InlineData(HostType.HttpListener, TransportType.LongPolling, MessageBusType.Default)]
-        [InlineData(HostType.HttpListener, TransportType.Websockets, MessageBusType.FakeMultiStream)]
-        [InlineData(HostType.HttpListener, TransportType.ServerSentEvents, MessageBusType.FakeMultiStream)]
-        [InlineData(HostType.HttpListener, TransportType.LongPolling, MessageBusType.FakeMultiStream)]
+        //[InlineData(HostType.HttpListener, TransportType.ServerSentEvents, MessageBusType.Default)]
```
davidfowl: I think its fine to nuke the coverage for all of these things.

anurse: I'm still waffling between wiping it out and making it conditional on some Environment Variable or something to enable us to still run the "maximum coverage" run before releases or something... but I'm not sure.

davidfowl:
Approved.. I think we need to re-enable some of those tests. But we have too many variations

PR #4196, opened 2018-09-10
.NET Client support for redirect responses

### LINKED ISSUE #4172: Add support for redirection negotiate responses
We want to add support for "redirection" responses from the initial negotiate request so that the client can be redirected to a different server.

There is already a `Url` field, though I don't know if the client uses it. Plus it is doubtful that the old client would send a new negotiation payload.

Either way, we need to increment the protocol version so that old clients produce a useful error. I suggest incrementing it to `2.0` in this case (it's currently `1.5`, `1.6` would be fine as long as the client has the right behavior). If the client does not explicitly indicate support for the new version, the server should report an error.

If we change the protocol version, we could now change the behavior of the `Url` field such that it worked similar to ASP.NET Core SignalR, and send a new negotiate payload to that URL if it is present (and remove it from responses where a redirection URL isn't being used).

* [x] JS Client - Done in https://github.com/SignalR/SignalR/pull/4186
* [ ] .NET Client

/cc @davidfowl 

Required for Azure SignalR Service support (https://github.com/Azure/azure-signalr/issues/101)

anurse:
Fixes #4172 

FYI @vicancy

This adds support for redirect responses to the .NET Client. It is WIP because I use the test code from #4193  and I need to rebase this and retarget to master once that's in.


src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -469,39 +473,80 @@ protected virtual string OnSending()
...
+                                        }
+
+                                        // Update the URL based on the redirect response and restart the negotiation
+                                        Url = negotiationResponse.RedirectUrl.ToString();
+
+                                        if (!string.IsNullOrEmpty(negotiationResponse.AccessToken))
+                                        {
+                                            // This will stomp on the current Authorization header, but that's by design.
+                                            // If the server specified a token, that is expected to overrule the token the client is currently using.
+                                            Headers["Authorization"] = $"Bearer {negotiationResponse.AccessToken}";
```
vicancy: The `AccessToken` should also be applied to other requests `/reconnect`, `/ping`, `/send`.. etc. And for `ws` and `sse` requests, it should be appended in the query string.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -469,39 +473,80 @@ protected virtual string OnSending()
...
+                                    if (negotiationResponse.ProtocolVersion.Equals("2.0") && !string.IsNullOrEmpty(negotiationResponse.RedirectUrl))
+                                    {
+                                        // Ensure the URL ends with a trailing slash.
+                                        if (!negotiationResponse.RedirectUrl.EndsWith("/"))
+                                        {
+                                            negotiationResponse.RedirectUrl += "/";
+                                        }
+
+                                        // Update the URL based on the redirect response and restart the negotiation
+                                        Url = negotiationResponse.RedirectUrl.ToString();
```
vicancy: ~~`ToString`~~?

test/Microsoft.AspNet.SignalR.FunctionalTests/Client/RedirectionFacts.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                using (var connection = CreateHubConnection(host, path: "/redirect-loop"))
+                {
+                    await Assert.ThrowsAsync<InvalidOperationException>(() => connection.Start(host.TransportFactory()));
+                }
+            }
+        }
+
+        [Fact]
+        public async Task DoesNotFollowRedirectIfProtocolVersionIsNot20()
```
BrennanConroy: Nit: 2_0?

test/Microsoft.AspNet.SignalR.FunctionalTests/Client/RedirectionFacts.cs
```diff
@@ -0,0 +1,86 @@
...
+        [Fact]
+        public async Task RedirectsAreLimitedToPreventInfiniteLooping()
+        {
+            using (var host = CreateHost(HostType.Memory, TransportType.Auto))
+            {
+                host.Initialize();
+
+                using (var connection = CreateHubConnection(host, path: "/redirect-loop"))
+                {
+                    await Assert.ThrowsAsync<InvalidOperationException>(() => connection.Start(host.TransportFactory()));
```
BrennanConroy: `OrTimeout` :trollface:

src/Microsoft.AspNet.SignalR.Client/Connection.cs
davidfowl: It doesn't need to be in the query string for any client but the javascript client.

anurse: Yep, we don't have the same limitations in .NET that we have in the browser. It's only the browser that forbids headers in EventSource and WebSocket requests.

test/Microsoft.AspNet.SignalR.FunctionalTests/Client/RedirectionFacts.cs
anurse: Meh.

test/Microsoft.AspNet.SignalR.FunctionalTests/Client/RedirectionFacts.cs
anurse: 💯

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -469,39 +473,80 @@ protected virtual string OnSending()
...
+
+                Monitor = new HeartbeatMonitor(this, _stateLock, beatInterval);
+
+                return StartTransport();
+            }
 
-                                    beatInterval = _keepAliveData.CheckInterval;
-                                }
-                                else
+            Task StartNegotiation()
```
anurse: `await` is for the weak 😸 

(but seriously: I can't change this to async/await unless we drop net40)

src/Microsoft.AspNet.SignalR.Client/Connection.cs
anurse: Whoops! Good catch. I had this factored to use `System.Uri` to "normalize" the URL but decided to just do the `.EndWith` thing above instead and forgot to remove this.

anurse:
Rebased and now ready to merge upon review completion

src/Microsoft.AspNet.SignalR.Client/Connection.cs
vicancy: Knowledge get!

PR #4199, opened 2018-09-14
clarify error when using JSONP or Forever Frame

anurse:
also added a sample that uses the Azure SignalR Service, bring your own connection string :).

/cc @vicancy


src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.foreverFrame.js
```diff
@@ -69,6 +69,14 @@
         iframeClearThreshold: 50,
 
         start: function (connection, onSuccess, onFailed) {
+            if (connection.accessToken) {
+                if (onFailed) {
+                    connection.log("Forever Frame does not support connections that require a Bearer token to connect, such as the Azure SignalR Service.");
```
BrennanConroy: Can this be a resource string like `jsonpNotSupportedWithAccessToken`?

anurse: I don't think it's as easy here since there are no existing resources in this file. Also I'm not sure we'll be localizing these messages anyway.

anurse: Also log messages aren't generally localized

samples/Microsoft.AspNet.SignalR.ServiceSample/packages.config
```diff
@@ -0,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="jQuery" version="1.6.4" targetFramework="net461" />
```
davidfowl: Holy crap 😄

anurse: .gif)

anurse:
This needs to be revisited. Things got a little out of date here. I'll work on a replacement.

PR #4210, opened 2018-09-20
chat client sample that works with the service sample

anurse:
This sample works with the Azure SignalR Service sample!

Usage:
1. Launch Microsoft.AspNet.SignalR.ServiceSample with an Azure SignalR connection string in the Web.config
2. `cd samples/MIcrosoft.AspNet.SignalR.Client.Samples`
3. `dotnet run -- chat --url [url]` where `[url]` is the `/signalr` endpoint for the service sample (`http://localhost:1234/signalr`, if the sample is running on `localhost:1234`)


anurse:
Adding this sample actually causes some problems since we end up in a weird reference loop because the service SDK depends on our packages. I think it would be better for samples to live with the Service (like in https://github.com/aspnet/AzureSignalR-samples)

PR #4220, opened 2018-09-25
fix #4175 by making TryGetConnectionId virtual

### LINKED ISSUE #4175: [SignalR Service Support] virtualize TryGetConnectionId for PersistentConnection
As per https://github.com/Azure/azure-signalr/pull/159#discussion_r210213550, currently we need to override `connectionToken` with `{connectionId}{userToken}` because there is check inside https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/PersistentConnection.cs#L324. @davidfowl suggests that `TryGetConnectionId` should just be made virtual so that we can avoid parsing the querystring

anurse:
The Azure SignalR Service SDK uses it's own custom connection tokens. Currently, make things work, they manually hack up a connection token that will validate correctly when they pass it through their PersistentConnection. This change allows any PersistentConnection to provide custom `connectionToken` validation and parsing logic by overriding the `TryGetConnectionId` method.

cc @vicancy

Fixes #4175

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/PersistentConnection.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/PersistentConnection.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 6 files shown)


davidfowl:
Can you write a test?

anurse:
What do you want to test here? The only change is that it's now virtual when it wasn't before.

davidfowl:
Lets just make sure you can override it and that it gets called.

anurse:
I'm working on a test, but pushed incomplete code so I can work on it at home ;). Will ping when it's ready to review.

anurse:
:up: :date: Ok, tests are in place now.

PR #4222, opened 2018-09-27
fix #4092 by disposing the response if it's a failed status code

### LINKED ISSUE #4092: The .NET 4.5 C# client leaks connections given responses with non-successful status code
### Expected  behavior

When a request made by the .NET 4.5 SignalR client fails due to a non-successful status code, the underlying connection should be gracefully closed by the time the SignalR client enters the disconnected state.

### Actual behavior

The underlying connection stays open because response was never disposed prior to throwing [here](https://github.com/SignalR/SignalR/blob/f5544f24c4a79bf2cca8850830a81bcfb67e1feb/src/Microsoft.AspNet.SignalR.Client45/Http/DefaultHttpClient.cs#L87) and [here.](https://github.com/SignalR/SignalR/blob/f5544f24c4a79bf2cca8850830a81bcfb67e1feb/src/Microsoft.AspNet.SignalR.Client45/Http/DefaultHttpClient.cs#L143) This can lead to a large number of open connections in combination with manual reconnect logic.

image

### Steps to reproduce

Repeatedly attempt to connect (i.e manual reconnect logic) to a SignalR endpoint that responds with a 401 from the /negotiate.

@davidfowl

anurse:
fixes #4092 by disposing the response on a failure.

TODO:
* [ ] Verify the full end-to-end behavior is fixed using repro for #4092

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Http/DefaultHttpClient.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Http/DefaultHttpClient.cs
@@ -86,7 +86,11 @@ namespace Microsoft.AspNet.SignalR.Client.Http
                      }
                      else
                      {
-                         throw new HttpClientException(responseMessage);
+                         // Dispose the response (https://github.com/SignalR/SignalR/issues/4092)
+                         var message = responseMessage.ToString();
+                         responseMessage.RequestMessage?.Dispose();
+                         responseMessage.Dispose();
+                         throw new HttpClientException(message);
                      }
 
                      return (IResponse)new HttpResponseMessageWrapper(responseMessage);
@@ -154,7 +158,7 @@ namespace Microsoft.AspNet.SignalR.Client.Http
         /// </summary>
         /// <param name="isLongRunning">Indicates whether the request is long running</param>
         /// <returns></returns>
-        private HttpClient GetHttpClient(bool isLongRunning)
+        private protected virtual HttpClient GetHttpClient(bool isLongRunning)
         {
             return isLongRunning ? _longRunningClient : _shortRunningClient;
         }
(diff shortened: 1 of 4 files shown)


BrennanConroy:
The issue points out https://github.com/SignalR/SignalR/blob/f5544f24c4a79bf2cca8850830a81bcfb67e1feb/src/Microsoft.AspNet.SignalR.Client45/Http/DefaultHttpClient.cs#L143 too

anurse:
Yep, I'll add a test for Post as well

src/Microsoft.AspNet.SignalR.Client/Http/DefaultHttpClient.cs
```diff
@@ -86,7 +86,11 @@ public Task<IResponse> Get(string url, Action<IRequest> prepareRequest, bool isL
                      }
                      else
                      {
-                         throw new HttpClientException(responseMessage);
+                         // Dispose the response (https://github.com/SignalR/SignalR/issues/4092)
+                         var message = responseMessage.ToString();
+                         responseMessage.RequestMessage.Dispose();
+                         responseMessage.Dispose();
+                         throw new HttpClientException(message);
```
davidfowl: Why are we passing in a string here?

halter73: So the client can see the failed response details if desired. The string was being used for the Exception message before. The only difference is instead of ToString() being called [here](https://github.com/SignalR/SignalR/blob/cf945d219a52c8103bde8bd5c6dc79f1ee0e64da/src/Microsoft.AspNet.SignalR.Client/Infrastructure/HttpClientException.cs#L46) like before, it's called before the HttpResponseMessage is disposed. Technically, I think ToString() still works after disposal, but it feels safer to just call ToString() earlier.

anurse: What @halter73 said :)

Issue #4075, opened 2018-02-13
XML parsing error: Hub reply does not define a root element 


kblok:
I'm replicating the [issue reported in the SignalR Core repo](https://github.com/aspnet/SignalR/issues/1106)

>This doesn't present a problem of the function, it is just not pretty; when invoking a method on the server my FF Developer tab reports:
>XML Parsing Error: no root element found
>Location: https://localhost:44311/signalr?id=ff8ff481-439a-4bcc-8f1f-67941a5ca331
>Line Number 1, Column 1:

>Again, not a bug per se but it does clutter the dev window. Not sure if it can be changed or if the browser is wrong.

I'm getting the same issue on FF when SignalR calls the close function. I'm getting this when the page is unloaded. It's not a big deal, but nobody like getting errors on the dev tools.

isaacplmann:
I'm experiencing the same issue.

Nath0s:
Having the same issue too...

electricessence:
Same here.  It seemed to not be this way not but yesterday.  Then I changed my hubs to reduce the number of methods.  In fact, eliminating all methods because only want to emit events.

I tried adding methods, to see if the XML changed, but no dice.

molo9966:
Using .NET MVC SIGNALR and jquery signalR 2.2.2 and, in firefox, getting following in console...

XML Parsing Error: no root element found
Location: http://l-mmoloney2.qvidiancorp.com/Qvidian.WOPI/signalr/signalr/abort?transport=we.......

This seems to then stop any SIGNALR messaging for the browser.  
Is there a new release upcoming to fix this?

omaximenco:
2.2.3 - same issue. It seems like SignalR for asp.net is a forgotten project...

electricessence:
@omaximenco It's definitely not.   It's in NPM without jQuery.  As the new Core version.
The issue is kinda jQuery because a request without a response body is fine, but only works without throwing in jQuery if you specify the content to be text and not json. :(

Nath0s:
@molo9966 Apparently, firefox is losing the connection right after the first connection attempt, the only fix i found was to call the connection function after the disconnection is detected
`//Only for firefox
    $.connection.hub.disconnected(function () {
        setTimeout(function () {
            $.connection.hub.start().done(function () {
            });
        }, 1);
    });`
With this it should be alright ;)

electricessence:
[...]

PR #4223, opened 2018-10-01
fix: "XML parsing error: Hub reply does not define a root element #4075"

lAnubisl:


src/Microsoft.AspNet.SignalR.Core/PersistentConnection.cs
```diff
@@ -154,7 +154,7 @@ public Task ProcessRequest(IDictionary<string, object> environment)
             environment.DisableResponseBuffering();
 
             var response = new OwinResponse(environment);
-
+            response.ContentType = "text/html";
```
anurse: Hrm, I think `text/html` is not a great choice for this. Also, we do set the mime type correctly in some requests (see `SendJsonResponse`) so I'd rather narrow this to make sure we understand where the Content-Type is missing and add the correct Content-Type values there.

lAnubisl: @anurse 
1) In this particular place if you call response.ContentType just after constructor call it will respond "text/html" but "Content-Type" will not be added to the response until you set it explicitly. Don't ask me why.
2) there are 2 places where ContentType is missing. But I think that all places which have a specific content type will just override default one. So I think that having default one is a good choice (you do want to have any response without Content-Type for any response don't you?)

anurse: > 1. Don't ask me why.

This is because it is a convention that `text/html` is the default Content-Type. `OwinResponse` is using that value to indicate that there is no active `Content-Type`. However, as you noticed, some browsers dislike that and emit a warning.

> 2\. So I think that having default one is a good choice

I don't think it's a good idea to have an inaccurate content type. I'd much prefer that we set it where we know what the content type is. Accidentally forgetting to set the content type yields an error message in the dev tools console which can easily be ignored and doesn't cause a funtional problem. Accidentally putting the wrong content type in place **can** yield functional problems. It's a matter of which is more dangerous to forget to do, and simply not setting the `Content-Type` is much less dangerous.

test/Microsoft.AspNet.SignalR.Tests/Server/Transports/ForeverTransportFacts.cs
```diff
@@ -94,6 +95,29 @@ public void AbortUrlTriggersConnectionAbort()
...
+            var hostContext = new HostContext(request.Object, response.Object);
+            var transportConnection = new Mock<ITransportConnection>();
+            var traceManager = new Mock<ITraceManager>();
+            var transport = new Mock<ForeverTransport>(hostContext, null, heartBeat.Object, counters.Object, traceManager.Object, null)
+            {
+                CallBase = true
+            };
+
+            transport.Object.ProcessRequest(transportConnection.Object).Wait();
+            response.VerifySet(r => r.ContentType = It.IsAny<string>(), "ContentType not set");
```
anurse: Could you assert the specific content type? We're trying to avoid simple "was this property set" or "was this method called" test cases moving forward since they are brittle and don't test the actual desired outcome.

src/Microsoft.AspNet.SignalR.Core/Transports/ForeverTransport.cs
```diff
@@ -106,6 +106,7 @@ protected Task ProcessRequestCore(ITransportConnection connection)
             }
             else if (IsAbortRequest)
             {
+                Context.Response.ContentType = "text/html";
```
anurse: Can we make this `text/plain`?

anurse:
Thanks for the contribution @lAnubisl !

PR #4228, opened 2018-10-04
Major deflaking

anurse:
I think I've removed most of the synchronous `Wait` calls from our tests so the thread-pool won't be starved during the test run. This should improve functional test reliability.


BrennanConroy:
Looks good except for all the NullRefs in tests

anurse:
:up: :date: fixed the test issue

PR #4229, opened 2018-10-05
Fix: synchronous XMLHttpRequest.

anurse:
Rebased version of #4025


anurse:
Thanks for the contribution @KyleScharnhorst !

PR #4231, opened 2018-10-08
add pdbs to nuget packages

### LINKED ISSUE #4109: Enable SourceLink for easier debugging
### Expected  behavior

Press F11 to step into SignalR code and debug the code

### Actual behavior

because no PDBs + source info are found, debugger steps over

### Solution: 

[enable SourceLink](https://github.com/dotnet/sourcelink#using-sourcelink) to [fetch the source from github](https://github.com/dotnet/designs/blob/master/accepted/diagnostics/source-link.md#source-link-file-specification) during debugging/step into to be able to step into the SignalR code.

anurse:
fixes #4109 by embedding the PDBs in the NuGet packages. See below for the impact on package sizes (all packages measured in Release builds):

| Package | Without (KB) | With (KB) | Increase |
|----------------------------------------------------------------|--------------|-----------|----------|
| Microsoft.AspNet.SignalR.Client| 242.34       | 325.64    | 34%      |
| Microsoft.AspNet.SignalR.Core| 160.60       | 213.52    | 33%      |
| Microsoft.AspNet.SignalR.Redis| 24.31        | 30.01     | 23%      |
| Microsoft.AspNet.SignalR.ServiceBus| 23.96        | 30.43     | 27%      |
| Microsoft.AspNet.SignalR.ServiceBus3| 24.93        | 31.50     | 26%      |
| Microsoft.AspNet.SignalR.SqlServer| 31.46        | 39.54     | 26%      |
| Microsoft.AspNet.SignalR.SystemWeb| 6.12         | 7.04      | 15%      |
| Microsoft.AspNet.SignalR.Utils| 19.53        | 24.14     | 24%      |
| **Total** | 533.25 | 701.81 | 32%|

If we need to use a symbol server I can investigate it, but this is a lot easier on the infrastructure and means users don't need to use a symbol server to download the PDBs (they're just always there). The percentages make it look a little high but given the improved debugging experience and the low absolute numbers (Redis + Core + SystemWeb = ~60KB total extra size; Client = ~85KB total extra size) I think this is the best way to do this.

+ @Eilon / @muratg in case they have any comments on shipping PDBs in the packages for **ASP.NET** SignalR

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -9,6 +9,9 @@
 
     <!-- Suppress missing documentation warnings -->
     <NoWarn>$(NoWarn);CS1591</NoWarn>
+
+    <!-- Embed PDBs in NuGet Packages -->
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
 
   <ItemGroup>


Eilon:
Are we really doing this instead of using a symbol server?

Eilon:
I'd rather we just have a ZIP people can download that has the PDBs. I don't want to start a trend here...

davidfowl:
Why though? This is simple and for older stuff that ships OOB. No Ngen, will always be packages, easy to maintain etc

Eilon:
This helps with local debugging but it doesn't easily help things like App Insights, does it? Unless you manually upload your PDBs to your deployed app, how will those services find the PDBs?

anurse:
@Eilon I don't mind how we do it, I'm just trying to find the right cost/benefit balance. Happy to start the investigation into a symbol server. Since this doesn't build as part of .NET Core we'll have to maintain all that infrastructure ourselves, but we can do it.

anurse:
I should note that nothing in this change prevents us from publishing to a symbol server at any point now or in the future since we can simply pick up the PDBs from the package and publish them.

Eilon:
BTW the reason I'm pushing a bit on this is that it took us like 3 years to come up with a solution that everyone agreed was the best option overall for PDBs in terms of format, where they are published, how they are consumed, etc., and that's why we arrived at source linking + symbol servers.

Have we looked at the cost of using some existing targets to do this push? Do we know how much work it really is? If it's a day, I think it's worth it, and then we can re-use that work on other projects (Web Stack, Katana, etc.).

muratg:
@Eilon I agree with your points for ASP.NET Core assemblies & packages. Do you feel as strongly for "classic" SignalR?

Eilon:
It just seems odd to me to make a different decision here. I'm not going to block it though.

anurse:
It looks like I can get the symbols published to the server without too much effort, so I'll look at doing that for preview and rtm releases.

I do agree with @davidfowl that the symbol server isn't always convenient so I prefer having both options available (PDBs in the NuPkgs and on the symbol server). Honestly, I'd prefer the debug information was just embedded in the DLL for all time. I can't think of a reason I'd ever not want it these days :)

karann-msft:
I came across this while I was searching for something else. BTW nuget.org symbol server is now a deafult option in VS which should alleviate some of those concerns about symbol server not being convenient for the user:

image

anurse:
We publish to the Microsoft Symbol Server now anyway, but thanks for the tip!

Issue #4207, opened 2018-09-20
Add "Error" property to Negotiate response


anurse:
To make it easier to for older clients to report errors due to incompatibilities, we should allow the `/negotiate` response to include an Error property. If present, the client will always cancel the connection and report the string provided as the error message.

anurse:
Among other errors, we should consider if reporting https://github.com/SignalR/SignalR/issues/4142 here makes sense.

PR #4234, opened 2018-10-08
fix #4142 by validating hub name in HubManager

### LINKED ISSUE #4142: Possible collision when HubName contains dot '.'
When I have 2 hubs:
```
 [HubName("advancedChat")]
    public class AdvancedChat1 : Hub {}
```
And
```
 [HubName("advancedChat.a")]
    public class AdvancedChat2 : Hub {}
```

When there is a Group `a.b` in Hub `advancedChat` and a Group `b` in Hub `advancedChat.a`, messages from the former Group will be received by the latter one.

### Expected  behavior
They should be separate.

### Actual behavior
They are not separate. The inner signal key `{HubName}.{GroupName}` collides. 

### Steps to reproduce
https://github.com/vicancy/SignalR/blob/lianwei/bugrepro/samples/SignalR2Chat2/AdvancedChat2.cs#L11 contains a sample project repro-ed the bug

anurse:
fixes #4142

This changes DefaultHubManager to throw if the hub name contains a '.',
since this is invalid and will conflict with the internal signal
"hc-{hubName}.{groupName}"

The exception manifests the following ways:
* signalr.exe generateHubProxies -> Outputs the error to the command
line
* signalr/js end-point -> YSOD with the error
* signalr/negotiate -> YSOD with error

When using the JS client, this manifests as a console error:

```
GET http://localhost:40476/signalr/signalr/negotiate?clientProtocol=2.0&connectionData=%5B%7B%22name%22%3A%22ch.at%22%7D%5D&_=1539032826063 500 (Internal Server Error)
```

Issue #4207 covers getting more detailed errors from negotiate. We can
evaluate exposing this from there when that comes online.


anurse:
cc @vicancy

src/Microsoft.AspNet.SignalR.Core/Hubs/Lookup/DefaultHubManager.cs
```diff
@@ -93,5 +101,13 @@ public IEnumerable<IHub> ResolveHubs()
         {
             return GetHubs(predicate: null).Select(hub => _activator.Create(hub));
         }
+
+        private void ValidateHubDescriptor(HubDescriptor hub)
+        {
+            if(hub.Name.Contains("."))
```
davidfowl: nit: formatting!

anurse:
> is it possible that the customer is not using `DefaultHubManager`?

Yes, but very unlikely and not officially supported. The HubManager is used in a number of places so I opted to put the validation in a central location. We could move it out but the risk is that we would fail to put the check in somewhere.

samples/Microsoft.AspNet.SignalR.Samples/Hubs/Chat/Chat.cs
```diff
@@ -38,7 +38,7 @@ public bool Join()
                 return false;
             }
 
-            ChatUser user = _users.Values.FirstOrDefault(u => u.Id == userIdCookie.Value);
+            var user = _users.Values.FirstOrDefault(u => u.Id == userIdCookie.Value);
```
anurse: The new VS "format document" action just does this automatically :).

PR #4245, opened 2018-10-12
fix #4207 by adding 'Error' to negotiate response

### LINKED ISSUE #4207: Add "Error" property to Negotiate response

anurse:
fixes #4207

This adds an "Error" field to the negotiate payload that contains an
error message provided by the server. If present, the `.start`/`.Start`
Task/Promise will be faulted with that error. This only works with
clients that support Protocol version 2.0, so it also has to have the
ProtocolVersion 2.0 value. This is designed to prevent older clients
from continuing to attempt to connect when they get the error response.
Since we use a new ProtocolVersion, older clients will fail with the
protocol incompatibility error.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Connection.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Connection.cs
@@ -511,35 +511,40 @@ namespace Microsoft.AspNet.SignalR.Client
                                 {
                                     VerifyProtocolVersion(negotiationResponse.ProtocolVersion);
 
-                                    if (negotiationResponse.ProtocolVersion.Equals("2.0") && !string.IsNullOrEmpty(negotiationResponse.RedirectUrl))
+                                    if (negotiationResponse.ProtocolVersion.Equals("2.0"))
                                     {
-                                        // Ensure the URL ends with a trailing slash.
-                                        if (!negotiationResponse.RedirectUrl.EndsWith("/"))
+                                        if (!string.IsNullOrEmpty(negotiationResponse.Error))
                                         {
-                                            negotiationResponse.RedirectUrl += "/";
+                                            // REVIEW: Exception type?
+                                            throw new Exception(string.Format(Resources.Error_ErrorFromServer, negotiationResponse.Error));
                                         }
-
-                                        // Update the URL based on the redirect response and restart the negotiation
-                                        Url = negotiationResponse.RedirectUrl;
-
-                                        if (!string.IsNullOrEmpty(negotiationResponse.AccessToken))
+                                        if (!string.IsNullOrEmpty(negotiationResponse.RedirectUrl))
                                         {
-                                            // This will stomp on the current Authorization header, but that's by design.
-                                            // If the server specified a token, that is expected to overrule the token the client is currently using.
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -511,35 +511,40 @@ Task StartNegotiation()
...
-                                    if (negotiationResponse.ProtocolVersion.Equals("2.0") && !string.IsNullOrEmpty(negotiationResponse.RedirectUrl))
+                                    if (negotiationResponse.ProtocolVersion.Equals("2.0"))
                                     {
-                                        // Ensure the URL ends with a trailing slash.
-                                        if (!negotiationResponse.RedirectUrl.EndsWith("/"))
+                                        if (!string.IsNullOrEmpty(negotiationResponse.Error))
                                         {
-                                            negotiationResponse.RedirectUrl += "/";
+                                            // REVIEW: Exception type?
+                                            throw new Exception(string.Format(Resources.Error_ErrorFromServer, negotiationResponse.Error));
```
mikaelm12: Hmm. Would `ExternalException` work here? I've never actually used it but found it in a quick search

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -729,86 +730,95 @@
...
-                                onFailed(signalR._.error(resources.errorRedirectionExceedsLimit), connection);
+                        if (res.ProtocolVersion === "2.0") {
+                            if (res.Error) {
+                                protocolError = signalR._.error(signalR._.format(resources.errorFromServer, res.Error));
+                                $(connection).triggerHandler(events.onError, [protocolError]);
+                                deferred.reject(protocolError);
                                 return;
                             }
+                            else if (res.RedirectUrl) {
+                                if (redirects === MAX_REDIRECTS) {
```
mikaelm12: nit: Don't think this could happen but it should be greater than or equal to

test/Microsoft.AspNet.SignalR.Tests.Common/App_Start/Initializer.cs
```diff
@@ -404,6 +426,7 @@ private static void RegisterSignalREndpoints(IAppBuilder app, IDependencyResolve
                 map.MapSignalR();
             });
 
+
```
mikaelm12: nit: clean this up

mikaelm12:
And rebase ofc

src/Microsoft.AspNet.SignalR.Client/Connection.cs
anurse: That exception is for COM Interop. We generally avoid throwing those.

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
anurse: We always increment by 1 so it shouldn't happen. Also this PR is just moving that code around, the code is as it was before.

PR #4247, opened 2018-10-15
fix #4152 by adding '/configFile' to ghp command

### LINKED ISSUE #4152: signalr.exe ghp crashes if project references NewTonSoft
If a web-project references NewtonSoft.Json then signalr.exe fails to generate proxy JS.

### Expected  behavior

js-file

### Actual behavior

Crash

``Error: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileLoadException: Could not load file or assembly 'Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)```

Temp workaround - manually creating a `signalr.exe.config` file and setting assembly redirect in the file like described [here](https://stackoverflow.com/a/33400441/56621). But this sucks cause /packages folder is usually not tracked by git, so it's a PITA to fix this for everyone in a big team.

FIX THIS PLEASE

anurse:
fixes #4152

This change adds support for a new '/configFile:' option to the 'signalr.exe' tool's 'ghp' command

If your server app uses a later version of an assembly than the 'signalr.exe' tool uses, we need a binding redirect. Newtonsoft.Json is the most common trigger here.

To fix this, you need to apply the binding redirect to the AppDomain that 'signalr.exe' creates to invoke the JavaScriptGenerator. This means you either need to put it in a 'signalr.exe.config' next to 'signalr.exe' (yuck), or we need to provide a different mechanism. This change makes it possible to use the '/configFile:' option to pass in the path to your app's Web.config file, which should already contain the necessary redirects.

Our own build tests this change! I updated the Samples project to JSON.NET 11 and immediately hit this issue. When building the Samples project we use `signalr.exe ghp` to generate hub proxies. With these changes, if the `/configFile` option is **not** specified, the build fails with this error message:

```
5>EXEC : error : Failed to load assembly 'Newtonsoft.Json, Version=6.0.0.0,
Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' referenced by the server. If your
application uses a newer version of this assembly, this error is likely due to a missing
or inaccurate binding redirect. If your server application has a 'Web.config' or
'[AppName].exe.config' file with binding redirects in it, provide the path to that file
with the '/configFile' parameter. If you have already provided a '/configFile' parameter,
make sure the config file it points to has a binding redirect for 'Newtonsoft.Json,
Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' within it. See
https://aka.ms/about-binding-redirects for more information on binding redirects.
```

I added the above error message to help guide users when they encounter this issue because it is not an obvious fix :).

Once the `/configFile` option is provided (see the Samples csproj for an example) the issue goes away!


samples/Microsoft.AspNet.SignalR.Samples/Microsoft.AspNet.SignalR.Samples.csproj
```diff
@@ -1,442 +1,441 @@
...
+        </Reference>
+        <Reference Include="Microsoft.Owin.Security, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+            <SpecificVersion>False</SpecificVersion>
+            <HintPath>..\..\packages\Microsoft.Owin.Security.3.0.0\lib\net45\Microsoft.Owin.Security.dll</HintPath>
+        </Reference>
+        <Reference Include="Microsoft.Owin.Security.Cookies, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+            <SpecificVersion>False</SpecificVersion>
+            <HintPath>..\..\packages\Microsoft.Owin.Security.Cookies.3.0.0\lib\net45\Microsoft.Owin.Security.Cookies.dll</HintPath>
+        </Reference>
+        <Reference Include="Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```
anurse: FYI, use "hide whitespace changes" this is mostly just indent changes, with a few small lines of real change

anurse:
Planning to merge this afternoon unless I hear otherwise. Want to get a 2.4.0 preview 2 release moving :)

PR #4248, opened 2018-10-15
fix #4126 by suppressing disconnect exception after logging

### LINKED ISSUE #4126: Unhandled exception in override of HubBase.OnDisconnected(bool) causes SignalR to leak memory
# Formal bug report

## Expected behavior

Allowing an exception to travel up the call stack from a hub's `OnDisconnected(bool)` override does not, in any circumstance, cause SignalR to retain internal buffers and/or meta data for a given connection indefinitely.

## Actual behavior

Allowing an exception to travel up the call stack from a hub's `OnDisconnected(bool)` override will cause SignalR to retain internal buffers and/or meta data for a given connection indefinitely if the client subscribed to events on the hub, therefore causing SignalR to invoke the connection lifecycle virtual methods of `HubBase`.

## Steps to reproduce

1. Throw an exception inside the `OnDisconnected(bool)` override of a hub
2. Override `HubBase.OnConnected` to trigger at least one thousand events with large arguments on the connected client, causing SignalR to fill its internal ring buffer for the connection with some bulky data (this is not done to cause the problem, merely to exagerate its effect so it can be more easily seen when profiling memory), then raise another event to signify when this process is finished
3. Create a proxy for the hub in the client and subscribe to the signal event, instructing the client to then disconnect by invoking the hub connection object's stop method or function
4. Execute the following test:
    1. Start the server and take a memory snapshot with the debugger or a profiler
    2. Start the client and let it run until finished
    3. Take a snapshot after the period at which the connection should have been ruled as disconnected and cleaned up

# Included example

[This ZIP archive](https://github.com/SignalR/SignalR/files/2060409/SignalRDisconnectTest.zip) contains a clone of the SignalR repository, but with disconnect test server and client projects added, as well as a new solution file that incorporates them. This was done so that it would be possible to step into SignalR and experiment with changes made to it (although this archive contains no such changes to SignalR). The server automatically forces a GC every second to make it possible to just watch memory usage over time without the developer forcing GC's via taking snapshots herself.

The server will ask the developer to select from among three testing modes:

* Normal: do not throw any exceptions in the `OnDisconnected(true)` override
* Throw: throw an exception in the `OnDisconnected(true)` override
* FaultedTask: return a faulted task from the `OnDisconnected(true)` override (such as what would occur if the method were decorated with async and the throw took place after an await)

Here are screenshots of the Visual Studio 2017 memory profiler displaying my results when running the server in each of these modes.

Normal:
normal

Throw:
throw

Faulted Task:
faultedtask

# A final word

I sincerely hope that no one accuses me of being lazy! ;)

I started with the intention of creating a PR to fix this problem. Unfortunately, finding exactly where to fix it elluded me, leaving aside the question of how. So, here I am, reporting what I know and providing the community with the test bench with which I replicated the issue in the hope that someone better than me can pick up where I left off and score a touchdown.

anurse:
Fixes #4126

If OnDisconnected throws or returns a faulted task, we don't dispose the MessageBus subscriptions associated with that connection. This results in the topic and associated messages leaking since nothing is left to dispose them.

This change suppresses the exception that arises in OnDisconnect after logging it so that the rest of the pipeline can carry on.

I haven't yet figured out a good way to test this. I did manual verification that there is indeed a leak in this scenario and that this change resolves the leak. cc @halter73 on this one as well since it's a bit obscure and I'd certainly value another set of SignalR-experienced 👀. It's only WIP because I might try to get a test figured out (though there isn't great existing coverage to base this on AFAICT) and I left some `// REVIEW` comments in place to illustrate the async flow for reviewers and I don't want them checked in.

TODO:
* [x] Remove `// REVIEW` comments put there to help reviewers understand the flow

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
--- a/src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
@@ -134,7 +134,6 @@ namespace Microsoft.AspNet.SignalR.Messaging
                 {
                     var messageResult = new MessageResult(items, totalCount);
 
-                    // REVIEW: If 'Invoke' faults, we fail to dispose and leak
                     bool result = await Invoke(messageResult, (s, o) => s.BeforeInvoke(o), state);
 
                     if (!result)
--- a/src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
@@ -134,6 +134,7 @@ namespace Microsoft.AspNet.SignalR.Messaging
                 {
                     var messageResult = new MessageResult(items, totalCount);
 
+                    // REVIEW: If 'Invoke' faults, we fail to dispose and leak
                     bool result = await Invoke(messageResult, (s, o) => s.BeforeInvoke(o), state);
 
                     if (!result)
--- a/src/Microsoft.AspNet.SignalR.Core/Transports/TransportDisconnectBase.cs
+++ b/src/Microsoft.AspNet.SignalR.Core/Transports/TransportDisconnectBase.cs
@@ -266,7 +266,7 @@ namespace Microsoft.AspNet.SignalR.Transports
             return Abort(clean: true);
         }
 
-        private async Task Abort(bool clean)
+        private Task Abort(bool clean)
         {
             if (clean)
             {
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
```diff
@@ -134,6 +134,7 @@ public async Task Work()
                 {
                     var messageResult = new MessageResult(items, totalCount);
 
+                    // REVIEW: If 'Invoke' faults, we fail to dispose and leak
```
davidfowl: Hmmmm, file a bug?

anurse: Yes, #4126 is that bug, and this is the fix.

I'm not 100% sure it's correct to catch the exception here but we could move it to there.

halter73: Given that Invoke isn't ever supposed to throw (It would get logged as an error in  MessageBroker.DoWork), do you think it best to defensively catch any exceptions and dispose the subscription?

src/Microsoft.AspNet.SignalR.Core/Transports/TransportDisconnectBase.cs
```diff
@@ -287,16 +287,22 @@ private Task Abort(bool clean)
             // End the connection
             End();
 
-            var disconnectTask = Disconnected != null ? Disconnected(clean) : TaskAsyncHelper.Empty;
-
             // Ensure delegate continues to use the C# Compiler static delegate caching optimization.
-            return disconnectTask
-                .Catch((ex, state) => OnDisconnectError(ex, state), state: Trace, traceSource: Trace)
```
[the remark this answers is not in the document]
halter73: This affects the OnDisconnected() overload too, right? I'm wondering why #4126 calls out the "hub's OnDisconnected(bool) override" specifically.

src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
anurse: Yeah, I wondered about that.

src/Microsoft.AspNet.SignalR.Core/Transports/TransportDisconnectBase.cs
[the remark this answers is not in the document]
anurse: There is only an `OnDisconnected(bool)` on `HubBase`

src/Microsoft.AspNet.SignalR.Core/Messaging/Subscription.cs
anurse: It's harder to log the error from here since we don't have an `ITraceManager` and adding one would be tricky. That's the main reason I left it to the callback to handle the exception.

halter73: I think calling `context.Subscription.Dispose()` in MessageBroker.DoWork's catch block is probably the way to go then. You'll have to make ISubscription implement IDisposable for that to work.

anurse: Yeah, I like that better. I can test it a bit there too.

anurse:
:up: :date: relocated the handling logic. Planning to merge this afternoon unless I hear otherwise :)

PR #4249, opened 2018-10-16
fix #4213 by tracking callback identity better

### LINKED ISSUE #4213: Hub callbacks add/remove issues
### Expected  behavior
Being able to add/remove callbacks properly.

### Actual behavior
- Callbacks get overwritten if the have the same string representation.
- Callbacks get executed even after removing them.

### Steps to reproduce
```
var fn1 = function() {console.log(1);};
var fn2 = function() {console.log(2);};
var fn3 = function() {console.log(3);};

// after this lines
// the string representation of fn1, fn2 and fn3 will be "function bound () { [native code] }"
fn1 = fn1.bind(this);
fn2 = fn2.bind(this);
fn3 = fn3.bind(this);

// suppose we have a hub with a 'change' event on it
hub.on('change', fn1);
hub.on('change', fn2);
hub.on('change', fn3);
hub.off('change', fn1);
```

the `fn1` will be still firing although `off` was called

### Investigation notes
In this line:
https://github.com/SignalR/SignalR/blob/master/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.hubs.js#L109

using callback as an index to an object triggers implicit toString invocation which from my experience returns the source code (if available, for native functions it just puts [native code] in the function body)

Very roughly said, for an event, it allows just one callback per function definition not per instance.

Did some more testing and it seems that the callbacks get overwritten in the callbackMap but still get registered with jQuery:
https://github.com/SignalR/SignalR/blob/master/src/Microsoft.AspNet.SignalR.Client.JS/jquery.signalR.hubs.js#L113

anurse:
fixes #4213

Previously we used the callback as the key in a JavaScript object. ECMAScript objects are required to have 'String' keys, which means the function is stringified and the string representation is used as the key. The problem is that functions can often have the same string representation. For example, if they are wrapped in something like `bind` or `apply`, the representation is `function() { [native code] }`.

This change switches to use an array and a 'guid' assigned by 'on' as the lookup system. That avoids the problems with stringifying functions. This system of assigning a 'guid' is used by jQuery internally for it's own event lists so it seems like the most appropriate way to handle this.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
@@ -7,6 +7,7 @@
 (function ($, window, undefined) {
     "use strict";
 
+    var nextGuid = 0;
     var eventNamespace = ".hubProxy",
         signalR = $.signalR;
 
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNet.SignalR.Client.JS.Tests/wwwroot/Tests/FunctionalTests/Hubs/HubEventHandlerFacts.js
```diff
@@ -61,6 +61,137 @@ testUtilities.runWithAllTransports(function (transport) {
...
+            },
+            handler2 = function (value) {
+                commonHandler(value, "handler2");
+            };
+
+        handler1 = handler1.bind(this);
+        handler2 = handler2.bind(this);
+
+        echoHub.on('echo', handler1);
+        echoHub.on('echo', handler2);
```
halter73: Any tests for manually providing a callbackIdentity?

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
```diff
@@ -7,6 +7,7 @@
 (function ($, window, undefined) {
     "use strict";
 
+    var nextGuid = 0;
```
halter73: guid is an interesting name for this

anurse: It's what jQuery uses... I'm not wedded to it at all.

test/Microsoft.AspNet.SignalR.Client.JS.Tests/wwwroot/Tests/FunctionalTests/Hubs/HubEventHandlerFacts.js
anurse: No, good catch, I'll do that. I added that feature after writing these tests ;P. Needed it for the `registerHubProxies` model.

anurse:
> Does this make `hub.off('foo', hub.client.foo)` work

Yes, it should work. I'll add a test.

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
```diff
@@ -91,56 +92,104 @@
...
+
+                    // Only unbind if there's an event bound with eventName and a callback with the specified callback
+                    if (callbackRegistration) {
+                        // Unbind all event handlers associated with the registration.
+                        for (var j = 0; j < callbackRegistration.eventHandlers.length; j++) {
+                            $(that).unbind(makeEventName(eventName), callbackRegistration.eventHandlers[j]);
+                        }
+
+                        // Remove the registration from the list
+                        callbackSpace.splice(i, 1);
```
Mythagoz: Surely this should be `callbackSpace.splice(callbackIndex , 1)` instead? Using `i` in the splice call will always do nothing as it is left at `callbackSpace.length` after the for loop above.
As this splice call will never empty the list the `delete callbackMap[eventName]` call below will never be made, potentially resulting in leaks.

PR #4277, opened 2018-12-14
npm audit fix

halter73:


test/Microsoft.AspNet.SignalR.Client.JS.Tests/package-lock.json
```diff
@@ -2251,9 +2256,9 @@
       "dev": true
     },
     "merge": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/merge/-/merge-1.2.0.tgz",
-      "integrity": "sha1-dTHjnUlJwoGma4xabgJl6LBYlNo=",
```
muratg: Ugh we didn't even have a direct dependency? :)

PR #4287, opened 2019-01-09
Fix NullRef in HeartbeatMonitor.Start()

### LINKED ISSUE #4285: System.NullReferenceException at Microsoft.AspNet.SignalR.Client.HeartbeatMonitor.Start()
# Issue:
Sometimes `System.NullReferenceException` throws `at Microsoft.AspNet.SignalR.Client.HeartbeatMonitor.Start() in /_/src/Microsoft.AspNet.SignalR.Client/HeartBeatMonitor.cs:line 59` when `Connection` fails quickly after `Start`.

# Root cause
1. `connection.Start(new DefaultHttpClient());` successfully and the websocket connection is established. web socket connection starts to process requests 
https://github.com/SignalR/SignalR/blob/2c9c833f40cd3c03fdc0de95099bff17118fee43/src/Microsoft.AspNet.SignalR.Client/Transports/WebSocketTransport.cs#L102
2. Server sends global error message to client, e.g. `{ "E": "Error from server" }`
3. Client tries to `connection.Disconnect()` with this global error: https://github.com/SignalR/SignalR/blob/2c9c833f40cd3c03fdc0de95099bff17118fee43/src/Microsoft.AspNet.SignalR.Client/Transports/ClientTransportBase.cs#L169-L173
4. In the meanwhile, `RunSynchronously` to `_transport.Start`, `Monitor.Start()` is about to run,
https://github.com/SignalR/SignalR/blob/2c9c833f40cd3c03fdc0de95099bff17118fee43/src/Microsoft.AspNet.SignalR.Client/Connection.cs#L571
5. `_connection` can be `null` if `connection.Disconnect()` is called before `Monitor.Start()`
https://github.com/SignalR/SignalR/blob/2c9c833f40cd3c03fdc0de95099bff17118fee43/src/Microsoft.AspNet.SignalR.Client/HeartBeatMonitor.cs#L59

halter73:
Fixes #4285

### DIFF
--- a/version.props
+++ b/version.props
@@ -8,5 +8,6 @@
     <FeatureBranchVersionPrefix Condition="'$(FeatureBranchVersionPrefix)' == ''">a-</FeatureBranchVersionPrefix>
     <VersionSuffix Condition="'$(VersionSuffix)' != '' And '$(FeatureBranchVersionSuffix)' != ''">$(FeatureBranchVersionPrefix)$(VersionSuffix)-$([System.Text.RegularExpressions.Regex]::Replace('$(FeatureBranchVersionSuffix)', '[^\w-]', '-'))</VersionSuffix>
     <VersionSuffix Condition="'$(VersionSuffix)' != '' And '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
+    <PackageVersion Condition="'$(PackageVersion)' == ''">$(VersionPrefix)-$(VersionSuffix)</PackageVersion>
   </PropertyGroup>
 </Project>


src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -558,17 +558,23 @@ private Task StartTransport()
...
+                                 {
+                                     // NOTE: We have tests that rely on this state change occuring *BEFORE* the start task is complete
+                                     if (!ChangeState(ConnectionState.Connecting, ConnectionState.Connected))
+                                     {
+                                         throw new StartException(Resources.Error_ConnectionCancelled, LastError);
+                                     }
+
+                                     // Now that we're connected complete the start task that the
+                                     // receive queue is waiting on
+                                     _startTcs.TrySetResult(null);
```
davidfowl: This tcs is dispatching to the thread pool right?

halter73: Correct. It's a "DispatchingTaskCompletionSource".

davidfowl: The change state also calls into user code right?

halter73: I checked both _startTcs and ChangeState() before submitting this PR to make sure we don't start calling user code under the _stateLock where we weren't before.

The StateChanged event *is* called indirectly by ChangeState(), but it's already being called under the lock: https://github.com/SignalR/SignalR/blob/2c9c833f40cd3c03fdc0de95099bff17118fee43/src/Microsoft.AspNet.SignalR.Client/Connection.cs#L398-L407.

Issue #4271, opened 2018-11-20
Microsoft.AspNet.SignalR.zh-Hans--Dependencies 2.4.0


ghd258:
Microsoft.AspNet.SignalR.zh-Hans--Dependencies 
Microsoft.AspNet.SignalR.SystemWeb.en-US (>= 2.4.0)   
（Oops!

Error 404 - Page Not Found

Looks like you might have gotten lost. Let's see if we can help you find your way!）

Microsoft.AspNet.SignalR.SystemWeb.zh-Hans (>= 2.3.0)
![1](https://user-images.githubusercontent.com/4811375/48753219-1aad9d00-ecc8-11e8-802c-ed89418e7cae.png)
![2](https://user-images.githubusercontent.com/4811375/48753225-200ae780-ecc8-11e8-9431-9198e6e77189.png)

anurse:
Weird, we'll look at fixing this.

anurse:
This is fixed in version [2.4.1](https://github.com/SignalR/SignalR/releases/tag/2.4.1) which is now released to NuGet.org and NPM.

PR #4297, opened 2019-01-23
Fix satellite dependency

halter73:
I've verified that this fixes nuget restore for projects that depend on Microsoft.AspNet.SignalR.$Locale$. "Microsoft.AspNet.SignalR.SystemWeb.en-US" has never existed, so I'm not sure how that got introduced. This should bring us back to parity with 2.3.0.

@davidfowl @muratg 	While this should fix the missing dependency issue, I think we should consider no longer publishing satellite packages unless we have a reasonable way to test it.

The experience just doesn't seem very good. For example, installing Microsoft.AspNet.SignalR.zh-Hans will bring in Microsoft.AspNet.SignalR.SystemWeb.zh-Hans, but it will not bring in Microsoft.AspNet.SignalR.Core.zh-Hans which seems pretty silly for a metapackage. This is because Microsoft.AspNet.SignalR.SystemWeb.zh-Hans depends on Microsoft.AspNet.SignalR.SystemWeb and that's it. There's no dependency on Core except for through the non-locale-specific SystemWeb package.

The SystemWeb satellite package seems particularly worthless since it doesn't define any resources. Of the 15 methods with localized intellisense docs in Microsoft.AspNet.SignalR.SystemWeb.xml, only GetHttpContext() still exists.

### DIFF
--- a/src/Microsoft.AspNet.SignalR/loc/Microsoft.AspNet.SignalR.satellite.nuspec
+++ b/src/Microsoft.AspNet.SignalR/loc/Microsoft.AspNet.SignalR.satellite.nuspec
@@ -16,7 +16,7 @@
         <tags>Microsoft AspNet SignalR AspNetSignalR websockets real-time realtime comet HTTP streaming</tags>
         <dependencies>
             <dependency id="Microsoft.AspNet.SignalR.JS" version="$PackageVersion$" />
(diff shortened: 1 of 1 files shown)


halter73:
Addresses #4271

davidfowl:
Why is the ci failing

halter73:
> Why is the ci failing

I hadn't merged https://github.com/SignalR/SignalR/pull/4287/commits/c928eeecb73ec505f469813d8678168e418117ee from the other PR.

Issue #4284, opened 2019-01-02
[Azure SignalR]Support query string for RedirectUrl for JS and C# client


vicancy:
When using Azure SignalR, it is possible that `/negotiate` returns `RedirectUrl` with query string inside, however, current JavaScript and C# client do not support such scenario

mikaelm12:
Hmm yeah, looks as though you are right. We just set the redirect url as the new negotiate url. In what scenario would the Azure SignalR Service return a redirect url with a query string? I don't think I've seen this before.
https://github.com/SignalR/SignalR/blob/fc475fa161eaf94150c4aaa4ba71830fd976eb6c/src/Microsoft.AspNet.SignalR.Client/Connection.cs#L520
https://github.com/SignalR/SignalR/blob/fc475fa161eaf94150c4aaa4ba71830fd976eb6c/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js#L749

vicancy:
Currently the core version of SignalR uses this to preserve original path and query string, see https://github.com/Azure/azure-signalr/issues/108#issuecomment-414355646 for one scenario. I am thinking if the ASP.NET one has similar requirements.

muratg:
Do we have any customer complaints? Is the broken scenario more of a corner case? Is there any workaround for customers? Trying to get a sense of patch-worthiness.

cc @bradygaster

bradygaster:
thanks for adding - i'd like to know if we've had many customer complaints on this before we dive in.

vicancy:
> Do we have any customer complaints that?

No, not yet. 

> Is there any workaround for customers

Yes, kind of, they can add the info to a custom claim.

muratg:
Thanks @vicancy. Let's see if we get any customer feedback on this to decide if we want to roll a patch.

vicancy:
Hi @muratg We now have one feature that dependent on this one https://github.com/Azure/azure-signalr/issues/346

muratg:
@vicancy thanks for the update. 

cc @davidfowl and @halter73 for their thoughts.

anurse:
This is fixed in version [2.4.1](https://github.com/SignalR/SignalR/releases/tag/2.4.1) which is now released to NuGet.org and NPM.

PR #4300, opened 2019-02-01
 Support query string for RedirectUrl for JS and C# client

halter73:
Addresses #4284

@vicancy

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Connection.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Connection.cs
@@ -93,6 +93,13 @@ namespace Microsoft.AspNet.SignalR.Client
 
         private readonly X509CertificateCollection _certCollection = new X509CertificateCollection();
 
+        // The query string passed to the ctor. Returned by Connection.QueryString.
+        private readonly string _userQueryString;
+
+        // The query string passed to the ctor plus what may have been added by the last
+        // RedirectUrl during negotiation.
+        private string _redirectPlusUserQueryString;
+
         // Keeps track of when the last keep alive from the server was received
         // internal virtual to allow mocking
         internal virtual HeartbeatMonitor Monitor { get; private set; }
@@ -174,7 +181,8 @@ namespace Microsoft.AspNet.SignalR.Client
             }
 
             Url = url;
-            QueryString = queryString;
+            _userQueryString = queryString;
+            _redirectPlusUserQueryString = queryString;
             _disconnectTimeoutOperation = DisposableAction.Empty;
             _lastMessageAt = DateTime.UtcNow;
             _lastActiveAt = DateTime.UtcNow;
@@ -374,7 +382,9 @@ namespace Microsoft.AspNet.SignalR.Client
         /// <summary>
         /// Gets the querystring specified in the ctor.
         /// </summary>
-        public string QueryString { get; private set; }
+        public string QueryString => _userQueryString;
+
+        string IConnection.QueryString => _redirectPlusUserQueryString;
 
         public IClientTransport Transport
         {
@@ -519,13 +529,33 @@ namespace Microsoft.AspNet.SignalR.Client
                                         }
                                         if (!string.IsNullOrEmpty(negotiationResponse.RedirectUrl))
                                         {
-                                            if (!negotiationResponse.RedirectUrl.EndsWith("/"))
+                                            try
                                             {
-                                                negotiationResponse.RedirectUrl += "/";
+                                                var redirectUrl = new UriBuilder(negotiationResponse.RedirectUrl);
+
+                                                if (!string.IsNullOrEmpty(redirectUrl.Query))
+                                                {
+                                                    var sb = new StringBuilder("?");
+                                                    UrlBuilder.AppendCustomQueryString(sb, redirectUrl.Query);
+                                                    UrlBuilder.AppendCustomQueryString(sb, _userQueryString);
+                                                    _redirectPlusUserQueryString = UrlBuilder.Trim(sb);
+                                                }
+
+                                                // Update the URL based on the redirect response and restart the negotiation
+                                                redirectUrl.Query = null;
+                                                Url = redirectUrl.ToString();
+                                            }
+                                            catch
+                                            {
+                                                // The most likely exception this catches is a UriFormatException, but we catch all exceptions and
+                                                // revert to the old behavior that doesn't attempt to extract the query string from the redirect URL.
+                                                Url = negotiationResponse.RedirectUrl;
                                             }
 
-                                            // Update the URL based on the redirect response and restart the negotiation
-                                            Url = negotiationResponse.RedirectUrl;
+                                            if (!Url.EndsWith("/"))
+                                            {
+                                                Url += "/";
+                                            }
 
                                             if (!string.IsNullOrEmpty(negotiationResponse.AccessToken))
                                             {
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -519,13 +529,33 @@ Task StartNegotiation()
                                         }
                                         if (!string.IsNullOrEmpty(negotiationResponse.RedirectUrl))
                                         {
-                                            if (!negotiationResponse.RedirectUrl.EndsWith("/"))
+                                            try
                                             {
-                                                negotiationResponse.RedirectUrl += "/";
+                                                var redirectUrl = new UriBuilder(negotiationResponse.RedirectUrl);
```
halter73: Do you think we should just split the string on '?' like we do for the JS client instead of attempting parsing it with UriBuilder?

I haven't come up with a case where it would work differently given a RedirectUrl that's valid enough to attempt a request with.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -519,13 +529,33 @@ Task StartNegotiation()
...
+
+                                                if (!string.IsNullOrEmpty(redirectUrl.Query))
+                                                {
+                                                    var sb = new StringBuilder("?");
+                                                    UrlBuilder.AppendCustomQueryString(sb, redirectUrl.Query);
+                                                    UrlBuilder.AppendCustomQueryString(sb, _userQueryString);
```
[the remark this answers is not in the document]
BrennanConroy: Ah, we keep the previous. Yep sounds good.

test/Microsoft.AspNet.SignalR.Client.JS.Tests/wwwroot/Tests/FunctionalTests/Core/RedirectFacts.js
```diff
@@ -25,6 +25,120 @@ QUnit.asyncTimeoutTest("Can connect to endpoint which produces a redirect respon
     });
 });
 
+QUnit.asyncTimeoutTest("Can connect to endpoint which produces a redirect response with a query string", testUtilities.defaultTestTimeout, function (end, assert, testName) {
+    // "/redirect-query-string2" -> "/signalr?name1=newValue&name3=value3"
+    var connection = testUtilities.createTestConnection(testName, end, assert, { url: "redirect-query-string2", hub: true }),
+        proxies = connection.createHubProxies(),
```
BrennanConroy: nit: formatting in this file, what's with the `,` instead of `;` and weird tabs?

halter73: I'm not sure what's weird about the tabs. Back in the old SignalR days, we typically used the comma operator when possible when declaring multiple variable at the top of a functions. When in Rome...

src/Microsoft.AspNet.SignalR.Client/Connection.cs
mikaelm12: This in most cases will only be happening once so creating the UriBuilder shouldn't have a big impact.

src/Microsoft.AspNet.SignalR.Client/Connection.cs
```diff
@@ -519,13 +529,29 @@ Task StartNegotiation()
...
+
+                                            if (splitUrlAndQuery.Length == 2 && !string.IsNullOrEmpty(splitUrlAndQuery[1]))
                                             {
-                                                negotiationResponse.RedirectUrl += "/";
+                                                var sb = new StringBuilder("?");
+                                                UrlBuilder.AppendCustomQueryString(sb, splitUrlAndQuery[1]);
+                                                UrlBuilder.AppendCustomQueryString(sb, _userQueryString);
+
+                                                // Update IConnection.QueryString with query string from only the most recent RedirectUrl.
+                                                _redirectPlusUserQueryString = UrlBuilder.Trim(sb);
```
halter73: I don't love how this preserves the last redirect query string if the client is restarted, but the same is true for the redirect url...

halter73:
I would like someone working on the SignalR service confirm that this change addresses 4284?
 
The change should make it possible to set an asrs.request.id token via the RedirectUrl query string in ProtocolVersion 2.0 negotiate response, but I have a few questions and design decisions that I would like feedback on:
 
1. Can you confirm that the only ‘?’ character in the RedirectUrl will be query string separator? This means any other instances of ‘?’ in the RedirectUrl need to be encoded.
2. I noticed while making this change that when the SignalR client restarts, it will continue to use the last RedirectUrl it received to instead of the URL it was initially configured with. After this change, the client will treat the query string included in the RedirectUrl similarly. Is this OK? Depending on how the service's redirect logic works, I’m worried might hinder the service's ability to change the query string when the SignalR client restarts.
3. I noticed that you referenced https://github.com/Azure/azure-signalr/issues/108 in https://github.com/SignalR/SignalR/issues/4284. I don’t think this PR helps with Azure/azure-signalr#108 since it only modifies the older clients which already had a mechanism for users to specify custom query strings. I want to verify we're on the same page here.
4. This PR merges the redirect query string with the user-specified query string. No attempt is made to avoid or detect conflicts, so it’s up to the service to choose query string keys that won’t conflict with user code.
5. If there are multiple redirects, the query string is only preserved from the last RedirectUrl.

vicancy:
> Can you confirm that the only ‘?’ character in the RedirectUrl will be query string separator? This means any other instances of ‘?’ in the RedirectUrl need to be encoded.

I think so. It makes things easier.

> I noticed while making this change that when the SignalR client restarts, it will continue to use the last RedirectUrl it received to instead of the URL it was initially configured with. After this change, the client will treat the query string included in the RedirectUrl similarly. Is this OK? Depending on how the service's redirect logic works, I’m worried might hinder the service's ability to change the query string when the SignalR client restarts.

Nice catch. Restarts with a new `/negotiate` towards the original URL can be helpful, especially when current Azure SignalR's token ttl is by default 1 hour. Restarts to the original URL can not only gain the ability to change query string but also renew the token. Do you think there is a need to improve current behavior?

> I noticed that you referenced [Azure/azure-signalr#108](https://github.com/Azure/azure-signalr/issues/108) in #4284. I don’t think this PR helps with [Azure/azure-signalr#108](https://github.com/Azure/azure-signalr/issues/108) since it only modifies the older clients which already had a mechanism for users to specify custom query strings. I want to verify we're on the same page here.

Not sure if I catch it. The ASP.NET Core one (the new client) should already support it. When client negotiates with app server, the SDK will preserve it's query string into the redirect URL, as:

https://github.com/Azure/azure-signalr/blob/617773638ea16dd1d838ae2cb2ab841fd50b90e7/src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs#L44-L45

> This PR merges the redirect query string with the user-specified query string. No attempt is made to avoid or detect conflicts, so it’s up to the service to choose query string keys that won’t conflict with user code.

Could we avoid doing the merge? When dealing with the new client, the original path and query string are actually captured and returned as query string of the redirect URL, as shown before. I'd like to keep it consistent with the old client. So is it possible to not merge but just trust the redirect URL?

> If there are multiple redirects, the query string is only preserved from the last RedirectUrl.

Sounds good, it keeps consistent with the new client. 

https://github.com/aspnet/SignalR/blob/8c6ed160d84f58ce8edf06a1c74221ddc8983ee9/clients/ts/signalr/src/HttpConnection.ts#L183

PR #4305, opened 2019-02-14
 Make client redirect logic more similar to the core clients

halter73:
@vicancy This is a follow up to #4300. The main changes are:

1. The user-specified query string is no longer merged with the redirect query string. The redirecting server is expected to reflect the user-specified query string.
2. The connection's URL and query string are now reset to their initial user-specified values when the connection stops instead of retaining their redirected values.
3. The .NET client's "Connection.Url" property no longer matches the redirect value and instead always maintains its initial value.
4. The JS clients "url" property does reflect the redirect value until the client stops, because the public url and related properties are used in various internal places. This is the already-existing behavior, and it does seem worthwhile to untangle it to behave like the .NET client.
5. Both the .NET clients "Connection.QueryString" property and the JS clients "qs" property never match the redirect query string and instead always maintain their initial value.

### DIFF
--- a/test/Microsoft.AspNet.SignalR.Client.JS.Tests/wwwroot/Tests/FunctionalTests/Core/RedirectFacts.js
+++ b/test/Microsoft.AspNet.SignalR.Client.JS.Tests/wwwroot/Tests/FunctionalTests/Core/RedirectFacts.js
@@ -25,6 +25,22 @@ QUnit.asyncTimeoutTest("Can connect to endpoint which produces a redirect respon
     });
 });
 
+QUnit.asyncTimeoutTest("Original URL used when client stops after a redirect response", testUtilities.defaultTestTimeout, function (end, assert, testName) {
+    var connection = testUtilities.createTestConnection(testName, end, assert, { url: "redirect", hub: true });
+
+    connection.start().done(function () {
+        assert.comment("Connection succeeded");
+        assert.equal(connection.url, window.location.protocol + "//" + window.location.host + "/signalr/", "The client redirected to the /signalr endpoint.");
+        connection.stop();
+        assert.equal(connection.url, "redirect", "The client's URL is reset to the initial value after stopping.");
+        assert.isNotSet(connection.baseUrl, "The client's URL-related state is reset after stopping.");
+        end();
+    }).fail(function () {
+        assert.fail("Connection failed!");
+        end();
+    });
+});
+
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNet.SignalR.Client.JS.Tests/wwwroot/Tests/FunctionalTests/Core/RedirectFacts.js
```diff
@@ -141,6 +156,39 @@ QUnit.asyncTimeoutTest("Clears redirect query string if last redirect doesn't ha
...
+        assert.comment("Connection succeeded");
+
+        hub.server.getQueryStringValue("name1").then(function (response) {
+            assert.equal(response, "newValue", "The client preserved 'name1=newValue' from redirect query string.");
+            return hub.server.getQueryStringValue("name3");
+        }).then(function (response) {
+            assert.equal(response, "value3", "The client preserved 'name3=value3' from redirect query string.");
+            return hub.server.getQueryStringValue("origName1");
+        }).then(function (response) {
+            assert.equal(response, "value1", "The client preserved 'name1=value1' from the first redirect query string for the next request in the redirect chain.");
```
BrennanConroy: nit: origName1

test/Microsoft.AspNet.SignalR.FunctionalTests/Client/NegotiateFacts.cs
```diff
@@ -60,30 +82,80 @@ public async Task CanMergeRedirectQueryStringWithUserQueryString()
...
+                    // Verify query string value set via (Hub)Connection ctor is absent. In practice, the redirect should reflect the query string.
+                    Assert.Equal(null, await hub.Invoke<string>("GetQueryStringValue", "foo"));
 
                     // Verify values set by redirect.
                     Assert.Equal("newValue", await hub.Invoke<string>("GetQueryStringValue", "name1"));
                     Assert.Equal("value3", await hub.Invoke<string>("GetQueryStringValue", "name3"));
 
                     // Verify that (Hub)Connection.QueryString only contains what was specified by the user, not the redirect.
-                    // IConnection.QueryString contains both, since this is what's used by the client to actually build URLs.
+                    // IConnection.QueryString contains what only what was set via redirect. This is what's used by the client to actually build URLs.
```
BrennanConroy: nit: remove extra 'what'

PR #4323, opened 2019-03-01
Make HubProxy state more thread-safe

### LINKED ISSUE #4319: HubProxy is not threadsafe when defining a State just before Invoking a method
before invoking a method I set a State:

image

and this piece of code is executed multiple times concurrently
### Expected  behavior
Works flawlessly
### Actual behavior
throws exception:

> at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeDictionary(JsonWriter writer, IDictionary values, JsonDictionaryContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)
   at Microsoft.AspNet.SignalR.Client.ConnectionExtensions.JsonSerializeObject(IConnection connection, Object value) in /_/src/Microsoft.AspNet.SignalR.Client/ConnectionExtensions.cs:line 53
   at Microsoft.AspNet.SignalR.Client.Hubs.HubProxy.Invoke[TResult,TProgress](String method, Action`1 onProgress, Object[] args) in /_/src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs:line 174
   at Microsoft.AspNet.SignalR.Client.Hubs.HubProxy.Invoke[T](String method, Object[] args) in src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs:line 76
**Result Message: System.InvalidOperationException : Collection was modified; enumeration operation may not execute.**

### Steps to reproduce
image

### Additional information:
In HubProxy the State is supposed to be threadsafe. Otherwise why would you put locks here:
image

But in invoke method you want to serialize the state(which involves enumeration of the dictionary) without a lock:

image

that gives "Collection was modified; enumeration operation may not execute." exception in a highly concurrent environment.
Please, fix it.

the solution could be to add an extra lock around serialization, or remove all locks and use the ConcurrentDictionary OR your best elegant solution

thanks!

halter73:
Fixes #4319

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs
@@ -166,12 +166,20 @@ namespace Microsoft.AspNet.SignalR.Client.Hubs
                 CallbackId = callbackId
             };
 
-            if (_state.Count != 0)
+            string value = null;
+
+            lock (_state)
             {
-                hubData.State = _state;
+                if (_state.Count != 0)
+                {
+                    hubData.State = _state;
+
+                    // Only keep the _state lock during JsonSerializeObject if hubData.State is set to _state.
+                    value = _connection.JsonSerializeObject(hubData);
+                }
             }
 
-            var value = _connection.JsonSerializeObject(hubData);
+            value = value ?? _connection.JsonSerializeObject(hubData);
 
             _connection.Send(value).ContinueWith(task =>
             {


src/Microsoft.AspNet.SignalR.Client/Hubs/HubProxy.cs
```diff
@@ -166,12 +166,20 @@ public Task Invoke<T>(string method, Action<T> onProgress, params object[] args)
...
+                {
+                    hubData.State = _state;
+
+                    // Only keep the _state lock during JsonSerializeObject if hubData.State is set to _state.
+                    value = _connection.JsonSerializeObject(hubData);
+                }
             }
 
-            var value = _connection.JsonSerializeObject(hubData);
+            value = value ?? _connection.JsonSerializeObject(hubData);
```
BrennanConroy: would there be any harm in putting all `_connection.JsonSerializeObject()` calls in the lock?
```csharp
lock (_state)
{
    if (_state.Count != 0)
    {
        hubData.State = _state;
    }
    value = _connection.JsonSerializeObject(hubData);
}
```

halter73: I don't think the lock will actually be contested in the vast majority of apps, but theoretically Invoke can be called from multiple threads, so I figure why take the risk of acquiring the lock for longer than necessary in the normal case when the state is empty.

BrennanConroy: After another look it doesn't make sense to do the other serialize inside the lock, I was thinking the `hubData.State` property could be set outside the lock, but it's a local.

PR #4329, opened 2019-03-06
Make invalid hub.off() calls from old generated hub proxies work

### LINKED ISSUE #4310: Js: hub methods get fired n+1 times on restarting connection
This new 'feature' came up while putting 2.4.0 in production a while ago. Users started getting duplicate or even sometimes 10x the same messages or any hub related events. In my code I have some reconnection logic as described in the [docs.](https://docs.microsoft.com/en-us/aspnet/signalr/overview/guide-to-the-api/handling-connection-lifetime-events#how-to-continuously-reconnect)

Turns out every time you invoke start, all hub methods get an additional subscription, causing all of them to get fired twice (or as many times as you invoke hub.start()).

### Expected  behavior

 hub.start() checks if subscriptions are in place and if so, doesn't add another one.

### Actual behavior

hub.start() adds extra subscription to all hub methods, causing them to fire multiple times.

### Steps to reproduce

install 2.4.0 js client: 
- start a simple signal chat program, with an js hub method onmessage or something. Disconnect signalr and start hub again.
- invoke hub method via a second browser, see message appearing twice

install 2.3.0 js client:
- do the same and see message appearing only 1 time.

halter73:
Fixes #4310

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
@@ -98,7 +98,9 @@
             /// <param name="callback" type="Function">The callback to be invoked.</param>
             /// <param name="callbackIdentity" type="Object">An optional object to use as the "identity" for the callback when checking if the handler has already been registered. Defaults to the value of 'callback' if not provided.</param>
             var that = this,
-                callbackMap = that._.callbackMap;
+                callbackMap = that._.callbackMap,
+                // https://github.com/SignalR/SignalR/issues/4310
+                isFromOldGeneratedHubProxy = !callbackIdentity && callback.toString().includes("// Call the client hub method");
 
             // We need the third "identity" argument because the registerHubProxies call made by signalr/js wraps the user-provided callback in a custom wrapper which breaks the identity comparison.
             // callbackIdentity allows the caller of `on` to provide a separate object to use as the "identity". `registerHubProxies` uses the original user callback as this identity object.
@@ -122,7 +124,7 @@
             // Check if there's already a registration
             var registration;
             for (var i = 0; i < callbackSpace.length; i++) {
-                if (callbackSpace[i].guid === callbackIdentity._signalRGuid) {
+                if (callbackSpace[i].guid === callbackIdentity._signalRGuid || (isFromOldGeneratedHubProxy && callbackSpace[i].isFromOldGeneratedHubProxy)) {
                     registration = callbackSpace[i];
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
```diff
@@ -98,7 +98,9 @@
             /// <param name="callback" type="Function">The callback to be invoked.</param>
             /// <param name="callbackIdentity" type="Object">An optional object to use as the "identity" for the callback when checking if the handler has already been registered. Defaults to the value of 'callback' if not provided.</param>
             var that = this,
-                callbackMap = that._.callbackMap;
+                callbackMap = that._.callbackMap,
+                // https://github.com/SignalR/SignalR/issues/4310
+                isFromOldGeneratedHubProxy = !callbackIdentity && callback.toString().includes("// Call the client hub method");
```
halter73: Yes, this is super gross. Generally, `hub.off` MUST be called with the same callback instance or callbackIdentity used during `hub.on`. The problem is that hub proxies generated by the SignalR server prior to 2.4.0 call `hub.on(eventName, callback)` and `hub.off(eventName, callback)` without a callbackIdentity and with *two different* callback instances.

We really should keep the old generated hub proxies working with the new client. The best way I've found so far to detect whether or not on/off is being called from the old generated hub proxy, is to look for the [// Call the client hub method](https://github.com/SignalR/SignalR/blob/d8f3843c17e382b238b64ade2548ead16f4d66d2/src/Microsoft.AspNet.SignalR.Core/Scripts/hubs.js#L24) comment from the generated function. If you know of any better alternatives, please tell me.

This wouldn't work if someone pre-generated the proxy using `signalr.exe ghp` and then minified it or something like that, but I'm not sure of any solution that handles that without completely reverting #4249.

This also means that there could be some issues if someone both uses the old generated hub proxy and manually calls `hub.on` for the same event using a function containing "// Call the client hub method", but I don't think that's a likely scenario that we should worry about.

halter73:
@davidfowl suggested that we revert #4249 and add two new methods to replace `hub.on` and `hub.off` that behaves the new way.

I understand wanting to do this considering the ugliness of this fix, but I doubt anyone would switch to using the 2 new on/off methods. There could also be some trickiness dealing callbacks that are added and removed using a combination of the new and old methods. At that point we might as well revert #4249, and be done.

Instead, I think we should just stick with this change, but I'd like to hear others feedback. @anurse

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
anurse: Ugh, server-owned client-side code is the W O R S T :)

Issue #4326, opened 2019-03-04
Update package license to be .NET Foundation / Apache


Eilon:
The package's license can match the source code license, so it should all just be .NET Foundation / Apache.

anurse:
This is fixed in version [2.4.1](https://github.com/SignalR/SignalR/releases/tag/2.4.1) which is now released to NuGet.org and NPM.

PR #4331, opened 2019-03-06
Update package license to be .NET Foundation / Apache

halter73:
This change links to the 2.3.0 version of LICENSE.txt. That file hasn't changed since 2.3.0. The previous non-official PackageLicenseUrl referenced the "master" branch, but it's generally considered bad practice to link to a living branch. In aspnet/AspNetCore, the PackageLicenseUrl always points to the 2.0.0 version of its LICENSE.txt, so there's precedent.

https://github.com/aspnet/AspNetCore/blob/e7b00a55083a56971e0d340ad2a3d350fec05883/Directory.Build.props#L14

Fixes: #4326

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -47,14 +47,14 @@
     <!-- Assembly/NuSpec attributes -->
     <Company>Microsoft</Company>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <PackageLicenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</PackageLicenseUrl>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/SignalR/SignalR/2.3.0/LICENSE.txt</PackageLicenseUrl>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(OfficialBuild)' != 'true'">
     <!-- Assembly/NuSpec attributes -->
     <Company>.NET Foundation</Company>
     <Copyright>© Copyright (c) .NET Foundation. All rights reserved.</Copyright>
-    <PackageLicenseUrl>https://raw.github.com/SignalR/SignalR/master/LICENSE.txt</PackageLicenseUrl>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/SignalR/SignalR/2.3.0/LICENSE.txt</PackageLicenseUrl>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


build/repo.targets
```diff
@@ -44,7 +44,6 @@
         <PropertyGroup>
             <_ConstantProperties>Configuration=$(Configuration)</_ConstantProperties>
             <_ConstantProperties>$(ConstantProperties);PackageVersion=$(PackageVersion)</_ConstantProperties>
-            <_ConstantProperties>$(ConstantProperties);LicenseUrl=http://example.com</_ConstantProperties>
```
halter73: Thankfully, looking at the [satellite packages on nuget.org](https://www.nuget.org/packages/Microsoft.AspNet.SignalR.Core.zh-Hans/2.4.0), it appears that http://example.com was never used as a LicenseUrl.

Prior to 2.4.0, the satellite package linked to localized satellite licenses. That's not longer the case, but I doubt it's a big issue. I don't think anyone else has even noticed.

Directory.Build.props
```diff
@@ -47,14 +47,14 @@
...
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <PackageLicenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</PackageLicenseUrl>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/SignalR/SignalR/2.3.0/LICENSE.txt</PackageLicenseUrl>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(OfficialBuild)' != 'true'">
     <!-- Assembly/NuSpec attributes -->
     <Company>.NET Foundation</Company>
     <Copyright>© Copyright (c) .NET Foundation. All rights reserved.</Copyright>
-    <PackageLicenseUrl>https://raw.github.com/SignalR/SignalR/master/LICENSE.txt</PackageLicenseUrl>
```
halter73: This change links to the 2.3.0 version of LICENSE.txt. In aspnet/AspNetCore, the PackageLicenseUrl always points to the 2.0.0 version of its LICENSE.txt, so there's precedent.

https://github.com/aspnet/AspNetCore/blob/e7b00a55083a56971e0d340ad2a3d350fec05883/Directory.Build.props#L14

Eilon:
Yup, all legit. Thanx @halter73 !

PR #4387, opened 2019-06-27
Use fetch in beforeunload

### LINKED ISSUE #4365: Latest Chrome will disable sync XHR calls in "beforeunload" please make according changes
Please see this bug: https://bugs.chromium.org/p/chromium/issues/detail?id=952452

Chrome update 73.0.3683.103 has broken the "beforeunload" event - synchronous XHR requests are no longer allowed from within this event. The Chrome team has reverted the change after reports from many software companies (see the bug discussion)

But the Chrome team **will** bring this back in June 2019 (again. see the bug discussion).

As far as I know, SignalR relies on "beforeunload" event to notify the server about a disconnect. Please change that. One option would be to switch from an ajax-call to the `sendBeacon` API.

alex-jitbit:
fixes #4365

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
@@ -448,15 +448,31 @@
 
             var url = getAjaxUrl(connection, "/abort");
             
-            if (navigator.sendBeacon) //use the new sendBeacon API cause new Chrome disabled XHR in "onbeforeunload"
+            var requestHeaders = connection.accessToken ? { "Authorization": "Bearer " + connection.accessToken } : {};
+            
+            //option #1 - try navigator.sendBeacon (reliable but unfortunately does not support Authorization header)
+            if (navigator.sendBeacon)
+            {
                 navigator.sendBeacon(url);
+            }
+            
+            //option #2 - send "fetch" with keepalive
+            if (window.fetch) //use the fetch API
+            {
+                fetch(url, {
+                    method: "POST",
+                    keepalive: true,
+                    headers: requestHeaders
+                });
+            }
 
+            //last resort - use plan old XHR (does not work in recent Chrome versions)
             transportLogic.ajax(connection, {
                 url: url,
                 async: async,
                 timeout: 1000,
                 type: "POST",
-                headers: connection.accessToken ? { "Authorization": "Bearer " + connection.accessToken } : {},
+                headers: requestHeaders,
                 dataType: "text" // We don't want to use JSONP here even when JSONP is enabled
             });
 
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
@@ -451,7 +451,8 @@
             var requestHeaders = connection.accessToken ? { "Authorization": "Bearer " + connection.accessToken } : {};
             
             //option #1 - send "fetch" with keepalive
-            if (window.fetch) { //use the fetch API
+            if (window.fetch) { 
+                // use the fetch API with keepalive
                 fetch(url, {
                     method: "POST",
                     keepalive: true,
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
@@ -459,7 +459,8 @@
                     headers: requestHeaders
                 });
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
```diff
@@ -447,6 +447,9 @@
             async = typeof async === "undefined" ? true : async;
 
             var url = getAjaxUrl(connection, "/abort");
+            
+            if (navigator.sendBeacon) //use the new sendBeacon API cause new Chrome disabled XHR in "onbeforeunload"
```
anurse: We put braces around all `if`'s regardless of size.

Also, the other code passes in an Authorization header, how do we do this in `sendBeacon`? I don't think it's possible to send headers, which is going to be blocking.

Another option could be to use [`fetch` with `keepAlive`](https://fetch.spec.whatwg.org/#ref-for-dom-request-keepalive) which would allow headers to work.

alex-jitbit: It **is** possible to send header with `sendbeacon`, `fetch` is unreliable in an `unload` that exactly why `sendbeacon` was invented - to send stuff when the windo is about to unload. I will make appropriate changes, hopefully it will take you less than 4 months to review the changes again

anurse: > I will add `fetch`, but its still unreliable in an `unload` that's exactly why `sendBeacon` was invented - to send stuff when the window is about to unload.

`fetch` **with the `keepAlive`** flag is designed to behave the same way as `sendBeacon`. In fact, Chrome implements `sendBeacon` using `fetch` with `keepAlive` now. So I think this is the appropriate change.

> hopefully it will take you less than 4 months to review the changes again ;)

Yep, apologies for the delay. We're the same team that works on ASP.NET Core and we were busy with the .NET Core 3.0 release. Feel free to ping us if we fall behind here!

alex-jitbit:
@anurse i made all the requested changes

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
```diff
@@ -447,15 +447,27 @@
...
-                timeout: 1000,
-                type: "POST",
-                headers: connection.accessToken ? { "Authorization": "Bearer " + connection.accessToken } : {},
-                dataType: "text" // We don't want to use JSONP here even when JSONP is enabled
-            });
+            
+            var requestHeaders = connection.accessToken ? { "Authorization": "Bearer " + connection.accessToken } : {};
+            
+            //option #1 - send "fetch" with keepalive
+            if (window.fetch) { //use the fetch API
```
anurse:
```suggestion
            if (window.fetch) { 
                // use the fetch API with keepalive
```

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.transports.common.js
```diff
@@ -447,15 +447,27 @@
...
+                fetch(url, {
+                    method: "POST",
+                    keepalive: true,
+                    headers: requestHeaders
+                });
+            }
+            else { //fetch no avail - fallback to tradiditonal $.ajax
```
anurse:
```suggestion
            else { 
                // fetch is not available - fallback to $.ajax
```

alex-jitbit:
Made all the changes, sorry it took so long, was out of town

crowmw:
```
  jquery.signalR.transports.common.js...
  Line 456 character 17: 'fetch' is not defined.
      fetch(url, {
```
gets this error while run `build.cmd`

alex-jitbit:
> gets this error while run `build.cmd`

I'm not aware of what `Build.cmd` does but after spending 20 minutes of my time looking at it I suppose it executes some JS-engine that has no support for `fetch`. Noda.js maybe?

This is really weird b/c literally 2 lines above this commsn I do the feature detection `if (window.fetch)`

alex-jitbit:
@anurse any news about including this?

anurse:
@alex-jitbit Sorry for the delay here, we need to do some testing here and haven't had a chance to prioritize it. I'll get this in to our next triage session and see if we can get someone to take a quick look this week. I don't think there's much to verify, but we should give it a cursory look :).

alex-jitbit:
It's been a year, come on now

halter73:
I tried this out locally on edge beta and chrome. Now when the tab is closed, fetch gets used with `keeepalive: true` to make the "/abort" request. `OnDisconnected(true)` still fires immediately as expected.

halter73:
Thanks @alex-jitbit

PR #4411, opened 2019-09-29
Update WebSocketTransport.cs

maxnis:
Change async void => async Task?


anurse:
If you're make this change, you'll need to update the call sites as well to ensure they discard the task (`_ = DoReconnect()`).

BrennanConroy (approved):

Issue #4419, opened 2019-10-14
Azure SQL Database managed instance not supported


norcino:
According with the [documentation](https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance), Azure SQL Database managed instances are compatible with SQL Server so it should be safe to add this to the supported DB targets.

### Expected  behavior
This database version (8) should be supported.

### Actual behavior
Exception is thrown on initialisation

### Steps to reproduce
Target a SQL Azure Managed Instance and run a signalR application

PR #4420, opened 2019-10-14
#4419 Added support for Sql Azure Managed Instance, which has an engine ID 8

norcino:
This changes allow signalR to be used with Sql Azure Managed Instances.

**Sql Azure Managed Instances** should be safely included in the supported DB Engines
Documentation reference: https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance

**Engine ID is 8**
Documentation reference: https://docs.microsoft.com/en-us/sql/t-sql/functions/serverproperty-transact-sql?redirectedfrom=MSDN&view=sql-server-ver15

**Testing**
I was unable to add tests to cover the changes because the updated class, **SqlInstaller.cs**, create instances of **DbOperation** which I cannot therefore mock.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.SqlServer/SqlInstaller.cs
+++ b/src/Microsoft.AspNet.SignalR.SqlServer/SqlInstaller.cs
@@ -53,7 +53,8 @@ namespace Microsoft.AspNet.SignalR.SqlServer
             var operation = new DbOperation(connectionString, "SELECT SERVERPROPERTY ( 'EngineEdition' )", _trace);
             var edition = (int)operation.ExecuteScalar();
 
-            return edition >= SqlEngineEdition.Standard && edition <= SqlEngineEdition.Express;
+            return edition >= SqlEngineEdition.Standard && edition <= SqlEngineEdition.Express ||
+                edition == SqlEngineEdition.SqlAzureManagedInstance;
         }
 
         private static class SqlEngineEdition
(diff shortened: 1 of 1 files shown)


anurse:
Looks good, thanks! Since Managed Instances have support for Service Broker we should be able to support them in the SQL backplane.

src/Microsoft.AspNet.SignalR.SqlServer/SqlInstaller.cs
```diff
@@ -53,7 +53,8 @@ private bool IsSqlEditionSupported(string connectionString)
             var operation = new DbOperation(connectionString, "SELECT SERVERPROPERTY ( 'EngineEdition' )", _trace);
             var edition = (int)operation.ExecuteScalar();
 
-            return edition >= SqlEngineEdition.Standard && edition <= SqlEngineEdition.Express;
+            return edition >= SqlEngineEdition.Standard && edition <= SqlEngineEdition.Express ||
```
anurse:
```suggestion
            return (edition >= SqlEngineEdition.Standard && edition <= SqlEngineEdition.Express) ||
```

I think the order of operations is correct without parens, but we like to be extra clear.

norcino: Changes applied as suggested, thank you.

PR #4436, opened 2019-12-18
Read TryWebSockets from last negotiate response in .NET client

### LINKED ISSUE #4435: net45 SignalR does not support WebSocket when using Azure SignalR
It's caused by:
https://github.com/SignalR/SignalR/blob/0ba22650e6eb1e4e4bf523c70951952a559db65d/src/Microsoft.AspNet.SignalR.Client/Transports/AutoTransport.cs#L77-L81, for `net45` version, it checks inside the redirect response from Azure SignalR that if it contains `TryWebSockets` in the redirect response which of cause not.

Easy to repro: Create a Console App under `net45`, install SignalR client package, and connect to app server which enables AzureSignalR. WebSocket transport is always excluded.

Is it a bug or by design?

halter73:
This fixes #4435 

#### Before:

```
02:22:03.0517617 - null - ChangeState(Disconnected, Connecting)
02:22:03.4398446 - PaDbxAX6F1J915s93MztvA4c7d46aa1 - SSE: GET https://notserverlessservice.service.signalr.net/aspnetclient/connect?clientProtocol=2.1&transport=serverSentEvents&connectionData=[{"Name":"ChatSampleHub"}]&connectionToken=PaDbxAX6F1J915s93MztvA4c7d46aa1&asrs_request_id=xEAfJaQDAAA%3D&asrs.op=%2Fsignalr
02:22:03.4807895 - PaDbxAX6F1J915s93MztvA4c7d46aa1 - SSE: OnMessage(Data: {"S":1,"M":[]})
02:22:03.5087902 - PaDbxAX6F1J915s93MztvA4c7d46aa1 - ChangeState(Connecting, Connected)
02:22:03.7990574 - PaDbxAX6F1J915s93MztvA4c7d46aa1 - SSE: OnMessage(Data: {})
```

#### After

```
02:23:45.3868953 - null - ChangeState(Disconnected, Connecting)
02:23:45.7129968 - pql9Bwbc_1lu9DJSKUKCpw4c7d46aa1 - WS Connecting to: wss://notserverlessservice.service.signalr.net/aspnetclient/connect?clientProtocol=2.1&transport=webSockets&connectionData=[{"Name":"ChatSampleHub"}]&connectionToken=pql9Bwbc_1lu9DJSKUKCpw4c7d46aa1&asrs_request_id=KoUbYqQDAAA%3D&asrs.op=%2Fsignalr
02:23:45.7609939 - pql9Bwbc_1lu9DJSKUKCpw4c7d46aa1 - WS: OnMessage({"S":1,"M":[]})
02:23:45.7899927 - pql9Bwbc_1lu9DJSKUKCpw4c7d46aa1 - ChangeState(Connecting, Connected)
02:23:45.7989927 - pql9Bwbc_1lu9DJSKUKCpw4c7d46aa1 - WS: OnMessage({})
```

@vicancy I see that https://github.com/Azure/azure-signalr/pull/774 has already made it to nuget.org. Impressive turnaround! I would keep that workaround in place indefinitely since you never know ho many pre-2.4.2 clients will stick around in the wild. Generally it's easier to update the server than it is to update every client.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Resources.Designer.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Resources.Designer.cs
@@ -187,6 +187,17 @@ namespace Microsoft.AspNet.SignalR.Client
         }
 
         /// <summary>
+        ///   Looks up a localized string similar to No client transport is compatible with the server. This either means that the AutoTransport was configured was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all..
+        /// </summary>
+        internal static string Error_NoCompatibleTransportFound
+        {
+            get
+            {
+                return ResourceManager.GetString("Error_NoCompatibleTransportFound", resourceCulture);
+            }
+        }
+
+        /// <summary>
         ///   Looks up a localized string similar to Possible deadlock detected. A callback registered with &quot;HubProxy.On&quot; or &quot;Connection.Received&quot; has been executing for at least {0} seconds..
         /// </summary>
         internal static string Error_PossibleDeadlockDetected
--- a/src/Microsoft.AspNet.SignalR.Client/Resources.Designer.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Resources.Designer.cs
@@ -187,7 +187,7 @@ namespace Microsoft.AspNet.SignalR.Client
         }
 
         /// <summary>
-        ///   Looks up a localized string similar to No client transport is compatible with the server. This either means that the AutoTransport was configured was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all..
+        ///   Looks up a localized string similar to No client transport is compatible with the server. This either means that the AutoTransport was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all..
         /// </summary>
         internal static string Error_NoCompatibleTransportFound
         {
--- a/src/Microsoft.AspNet.SignalR.Client/Resources.resx
+++ b/src/Microsoft.AspNet.SignalR.Client/Resources.resx
@@ -195,4 +195,7 @@
   <data name="Error_AspNetCoreServerDetected" xml:space="preserve">
     <value>Detected a connection attempt to an ASP.NET Core SignalR Server. This client only supports connecting to an ASP.NET SignalR Server. See https://aka.ms/signalr-core-differences for details.</value>
   </data>
+  <data name="Error_NoCompatibleTransportFound" xml:space="preserve">
+    <value>No client transport is compatible with the server. This either means that the AutoTransport was configured was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all.</value>
+  </data>
 </root>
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.SignalR.Client/Transports/AutoTransport.cs
```diff
@@ -74,14 +73,10 @@ public string Name
         public Task<NegotiationResponse> Negotiate(IConnection connection, string connectionData)
         {
             var task = GetNegotiateResponse(connection, connectionData);
-#if NET45
+#if NET45 || NETSTANDARD2_0
```
halter73: The missing `|| NETSTANDARD2_0` is why clients running on .NET Core could still still connect to the service with WebSockets. .NET Core clients would always try WebSockets regardless of the negotiate response 😞. I checked the rest of the codebase. This is the only place that `#if NET45` appeared by itself without some `|| NETSTANDARD...` component.

vicancy: How about checking if the negotiate response is redirect response or not to determine if `webSockets` should be removed?

halter73: Is there some reason that's preferable to using the last negotiate response? Because of the layering, I'd rather the AutoTransport not have to determine whether a negotiate response is a redirect or not. It's the code that's calling IClientTransport.Negotiate() that determines this today.

The better layering would be to have the Connection.Negotiate() (which can call the transport's negotiate method multiple times to follow redirects) look at the TryWebSockets value on the last negotiate response and feed the TryWebSockets value back into the transport as a parameter to IClientTransport.Start(). I didn't do that, because I didn't want to have to do a major refactoring and change the IClientTransport interface.

src/Microsoft.AspNet.SignalR.Client/Resources.resx
```diff
@@ -195,4 +195,7 @@
   <data name="Error_AspNetCoreServerDetected" xml:space="preserve">
     <value>Detected a connection attempt to an ASP.NET Core SignalR Server. This client only supports connecting to an ASP.NET SignalR Server. See https://aka.ms/signalr-core-differences for details.</value>
   </data>
+  <data name="Error_NoCompatibleTransportFound" xml:space="preserve">
+    <value>No client transport is compatible with the server. This either means that the AutoTransport was configured was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all.</value>
```
BrennanConroy:
```suggestion
    <value>No client transport is compatible with the server. This either means that the AutoTransport was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all.</value>
```

src/Microsoft.AspNet.SignalR.Client/Resources.Designer.cs
```diff
@@ -186,6 +186,17 @@ internal static string Error_NegotiationLimitExceeded
             }
         }
 
+        /// <summary>
+        ///   Looks up a localized string similar to No client transport is compatible with the server. This either means that the AutoTransport was configured was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all..
```
BrennanConroy:
```suggestion
        ///   Looks up a localized string similar to No client transport is compatible with the server. This either means that the AutoTransport was configured to use only WebSockets which is not compatible with the server or that the AutoTransport was configured with no sub-transports at all..
```

PR #4438, opened 2019-12-20
Fall back or reconnect but not both (.NET client)

halter73:
There was a race observed in the .NET client's ServerSentEventsTransport that caused the client to both try to fall back from SSE to long polling, and also reconnect with SSE. The following server logs show the buggy behavior.

UTC time | connection Id | Event | Method | URL
-- | -- | -- | -- | --
2019-12-02T00:21:12.2371847Z | foo | Request End | GET | http://ne-1.service.signalr.net/aspnetclient/connect?clientProtocol=2.1&transport=serverSentEvents&connectionData=[%7B%22Name%22:%22SomeHub%22%7D]&connectionToken=bar&asrs_request_id=baz%3D&asrs.op=%2Fsignalr
2019-12-02T00:21:12.2372764Z | foo | Request Start | POST | http://ne-1.service.signalr.net/aspnetclient/connect?clientProtocol=2.1&transport=longPolling&connectionData=[%7B%22Name%22:%22SomeHub%22%7D]&connectionToken=bar&asrs_request_id=baz%3D&asrs.op=%2Fsignalr
2019-12-02T00:21:14.2524252Z | foo | Request Start | GET | http://ne-1.service.signalr.net/aspnetclient/reconnect?clientProtocol=2.1&transport=serverSentEvents&connectionData=[%7B%22Name%22:%2222SomeHub%22%7D]&connectionToken=bar&asrs_request_id=baz%3D&asrs.op=%2Fsignalr

The following is what I believe happened. In the list I talk about the SSE transport, but since all the client transports share a lot of logic in ClientTransportBase, I believe this bug could have happened with any of the client's three transports.

1. The transport connects with SSE but doesn't get an init message before the TransportInitializationHandler times out IClientTransport.Start() causing it to return a faulted task which in turn causes the AutoTransport to fall back to long polling.
2. At about exactly the same time, the server closes the SSE /connect response with a FIN (this can be seen in the logs).
3. The SSE transport's EventSourceStreamReader.Closed handler runs.

Normally, Start() timing out would result in the SSE transports _stop flag being set in OnStartFailed() which would cause the Closed handler not to reconnect. Unfortunately, the TransportInitializationHandler only calls OnStartFailed() on a background thread after the decision to fault the Start() task has been made. This means there's a window right after the TransportInitializationHandler decides to timeout Start() where the _stop flag is still not set allowing the SSE transport to both reconnect and return a faulted Start() task.

The fix is for each transport to always try to fault IClientTransport.Start() if there's a transport disconnect/error. Then, if Start() was faulted for any reason (perhaps it was faulted even before this due to a timeout), don't try to reconnect. If the AutoTransport is in use, the faulted Start() results in a transport fallback, so not reconnecting is crucial.

@vicancy

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Transports/LongPollingTransport.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Transports/LongPollingTransport.cs
@@ -208,7 +208,8 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
                 TryReconnect(connection, _reconnectInvoker);
             }
 
-            if (shouldReconnect)
+            // Don't reconnect if Start() never succeeded.
+            if (shouldReconnect && !TryFailStart(new IOException(Resources.Error_TransportDisconnectedBeforeConnectionFullyInitialized)))
             {
                 // Transition into reconnecting state
                 connection.EnsureReconnecting();
@@ -239,16 +240,23 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
         // internal virtual for testing
         internal virtual void OnError(IConnection connection, Exception exception)
         {
-            TransportFailed(exception);
+            // Prevent the reconnecting -> connected transition from happening if it hasn't already.
             _reconnectInvoker.Invoke();
+
+            if (TryFailStart(exception))
+            {
+                return;
+            }
 
             if (!TransportHelper.VerifyLastActive(connection))
             {
                 StopPolling();
             }
-
-            // Transition into reconnecting state
-            connection.EnsureReconnecting();
+            else
+            {
+                // Do the connected -> reconnecting transition if it hasn't already occurred.
+                connection.EnsureReconnecting();
+            }
 
             // Sometimes a connection might have been closed by the server before we get to write anything
             // so just try again and raise OnError.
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.SignalR.Client/Transports/LongPollingTransport.cs
```diff
@@ -239,16 +240,23 @@ private Task OnAfterPoll(Exception exception)
...
+            }
+
             if (!TransportHelper.VerifyLastActive(connection))
             {
                 StopPolling();
             }
-
-            // Transition into reconnecting state
-            connection.EnsureReconnecting();
+            else
```
BrennanConroy: Why is this changing?

halter73: When VerifyLastActive returns false it should move to the disconnected state which should cause EnsureReconnecting to no-op. I was just making it more explicit in an attempt to make this code slightly easier to reason about.

BrennanConroy: Alright cool, I figured it was something like that, just checking

PR #4439, opened 2020-01-03
Avoid multiple simultanious DoReconnect loops

### LINKED ISSUE #4412: SignalR client leaks connections when server is down
I am hosting an AspNet SignalR Client (Microsoft.AspNet.SignalR.Client.dll v2.4.1) in an Azure AppService app. It connects to a SignalR server that I do not control. This server sometimes reboots or are down for maintenance. When this happens the number of outgoing connections quickly hits the limit of an AppService app (>1000). The connections are not freed up until I reboot the app. See screenshot of connections. 
signalr_connections

Is there a way for me to handle this scenario or is it an issue with the aspnet signalr client?

Log attached (I have removed tokens etc) 
[signalrclient.txt](https://github.com/SignalR/SignalR/files/3669416/signalrclient.txt)

halter73:
Fixes #4412

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Client/Transports/WebSocketTransport.cs
+++ b/src/Microsoft.AspNet.SignalR.Client/Transports/WebSocketTransport.cs
@@ -64,7 +64,7 @@ namespace Microsoft.AspNet.SignalR.Client.Transports
             _connectionData = connectionData;
 
             // We don't need to await this task
-            PerformConnect().ContinueWith(task =>
+            ConnectAndHandleConnection().ContinueWith(task =>
             {
                 if (task.IsFaulted)
                 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.SignalR.Client/Transports/WebSocketTransport.cs
```diff
@@ -166,13 +175,14 @@ internal virtual void OnClose()
...
         {
+            var token = CreateLinkedCancellationToken();
             var reconnectUrl = UrlBuilder.BuildReconnect(_connection, Name, _connectionData);
 
             while (TransportHelper.VerifyLastActive(_connection) && _connection.EnsureReconnecting())
             {
                 try
                 {
-                    await PerformConnect(reconnectUrl);
+                    await PerformConnect(reconnectUrl, token);
```
BrennanConroy: My one concern is that we're using the same `_webSocketTokenSource` in this loop whereas before it would be reset everytime. Do you know if that is ok? If `LostConnection` is called does this loop matter anymore?

halter73: You're probably right that it's better to move `var token = CreateLinkedCancellationToken();` inside the loop to avoid a possible regression.

The only non-terminal cancellation of _webSocketTokenSource is in LostConnection() which I don't think can happen until after the connection is reconnected, so I was trying to fix a possible race with dispose here by moving the creation of the _webSocketTokenSource outside the loop, but that's not the bug I was originally trying to fix with this PR.

PR #4459, opened 2020-04-07
Change official build pool

halter73:
@anurse @davidfowl 

@mmitche The [signalr-build.yml template](https://github.com/SignalR/SignalR/blob/cb6c9a91e9771d1bc7748f8ae1823fb29ed81a92/.vsts-pipelines/templates/signalr-build.yml) appears to be pretty outdated and doesn't use arcade's eng/common/templates/jobs/jobs.yml like dotnet/extensions does. Do I need to specify "queue: BuildPool.Server.Amd64.VS2019" like dotnet/extensions does [here](https://github.com/dotnet/extensions/blob/75759edc975749e1a8c4198c0078755a1822eac8/azure-pipelines.yml#L83)?

### DIFF
--- a/.vsts-pipelines/builds/ci-official.yml
+++ b/.vsts-pipelines/builds/ci-official.yml
@@ -9,7 +9,7 @@ trigger:
 phases:
 - template: ../templates/signalr-build.yml
   parameters:
-    queueName: DotNetCore-Windows
+    pool: internal
     variables:
       SignType: real
       Localize: true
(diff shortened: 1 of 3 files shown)


anurse:
> The [signalr-build.yml template](https://github.com/SignalR/SignalR/blob/cb6c9a91e9771d1bc7748f8ae1823fb29ed81a92/.vsts-pipelines/templates/signalr-build.yml) appears to be pretty outdated and doesn't use arcade's eng/common/templates/jobs/jobs.yml like dotnet/extensions does.

This repo doesn't use arcade period :). It probably should be updated, but it's not part of the .NET Core dependency flow at all (neither in nor out) so it's generally been fine as-is.

halter73:
Okay. I updated this to use 'pool' in stead of 'queue' syntax, so I can specify "BuildPool.Server.Amd64.VS2019".

PR #4464, opened 2020-04-13
Fix build

halter73:


.vsts-pipelines/builds/ci-official.yml
```diff
@@ -38,15 +42,16 @@ phases:
...
-    - task: PublishSymbols@2
-      displayName: 'Publish symbols path'
-      condition: succeeded()
-      inputs:
-        SearchPattern: 'artifacts\$(BuildConfiguration)\symbols\windows\**\*.pdb'
-        SymbolServerType: FileShare
-        SymbolsProduct: AspNetSignalR
-        SymbolsPath: '$(DropShareRoot)\$(Build.BuildNumber)'
-        SymbolsArtifactName: 'symbols-$(BuildConfiguration)'
+    # Temporarily disable the PublishSymbols task since the agent doesn't currently have access to the $(DropShareRoot)
```
halter73, line 45: REVIEW: Should we just remove this? It looks like the symbol packages are still accessible from the published artifacts.

anurse: If we remove this we'll have to publish the symbols manually for official releases. I think we can handle that, we just have to add it to the instructions in https://github.com/dotnet/aspnetcore-internal/blob/master/docs/release/AspNetSignalRBuildAndRelease.md#publish-a-release

Issue #4444, opened 2020-01-28
SQL Backplane - plan cache becomes bloated when sending due to @Payload parameter changing size


mak0t0san:
### Expected  behavior
SQL should just cache one plan. The `@Payload` parameter should be `varbinary(max)` instead of `varbinary(nnn)` {where nnn  changes based on the size of the message}.

### Actual behavior
The issue is in SqlSender when it creates the `@Payload` parameter.  It's a varbinary and when the SQL runs it will execute with different sizes (`@Payload varbinary(4001)`, `@Payload varbinary(4731)`, `@Payload varbinary(5000)`, etc.).  This causes SQL to cache each one of these separately and bloating the plan cache.
 
![image](https://user-images.githubusercontent.com/987722/73278396-85eedb00-41b9-11ea-9416-9c932cea55d7.png)

I've created a fork and made changes to fix the issue:
https://github.com/mak0t0san/SignalR/commit/f9b12cfcc269a71aeedbbfba00fc4351cd65e9e8

### Steps to reproduce
Use SignalR with SQL backplane.

laurihelkkula:
We are also seeing this problem. Transitioning to some other backplane would be non-trivial, so a fix for this would be nice. 

Would you take a PR for this? The change seems to pretty simple and I think I can also create a unit test.
I do have one question: Is it okay to use varbinary(max) for all payload sizes or is there some benefit for using one size (or few sizes) for payloads under that size and max for anything over that size? For example, use the varbinary(8000) for payloads under 8000 bytes and max for everything over 8000 bytes?  8000 is the max size for non-max varbinary.

BrennanConroy:
https://github.com/SignalR/SignalR/pull/4467#issuecomment-781662759

PR #4467, opened 2020-04-30
Change Payload parameter value to be a stream instead of a byte array

mak0t0san:
Change Payload parameter value to be a stream instead of a byte array
- The Payload SQL parameter will be `varbinary(max)` instead of `varbinary(nnn)` {where nnn is a value that changes based on the size of the messages}. This change will prevent the cache plan in SQL from becoming bloated.
- By using a stream type instead of a byte array, the resulting SQL statement will use the type `varbinary(max)` instead of `varbinary(nnn)` 

#4444

### DIFF
--- a/src/Microsoft.AspNet.SignalR.SqlServer/SqlSender.cs
+++ b/src/Microsoft.AspNet.SignalR.SqlServer/SqlSender.cs
@@ -4,10 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
-using System.Data.SqlClient;
-using System.Data.SqlTypes;
 using System.Diagnostics;
 using System.Globalization;
+using System.IO;
 using System.Threading.Tasks;
 using Microsoft.AspNet.SignalR.Messaging;
 
@@ -43,14 +42,17 @@ namespace Microsoft.AspNet.SignalR.SqlServer
                 return TaskAsyncHelper.Empty;
             }
 
-            var parameter = _dbProviderFactory.CreateParameter();
-            parameter.ParameterName = "Payload";
-            parameter.DbType = DbType.Binary;
-            parameter.Value = SqlPayload.ToBytes(messages);
+            using (var stream = new MemoryStream(SqlPayload.ToBytes(messages)))
+            {
+                var parameter = _dbProviderFactory.CreateParameter();
+                parameter.ParameterName = "Payload";
+                parameter.DbType = DbType.Binary;
+                parameter.Value = stream;
 
-            var operation = new DbOperation(_connectionString, _insertDml, _trace, parameter);
+                var operation = new DbOperation(_connectionString, _insertDml, _trace, parameter);
 
-            return operation.ExecuteNonQueryAsync();
+                return operation.ExecuteNonQueryAsync();
+            }
         }
     }
 }


davidfowl:
This looks good to me. AFAIK We don't have verification tests for the sql backplane though so this needs to be verified manually.

halter73:
I'm not a SQL expert, but it seems that setting the actual length is what's expected for entries smaller than 8000 bytes. https://docs.microsoft.com/en-us/sql/t-sql/data-types/binary-and-varbinary-transact-sql?view=sql-server-ver15

What's the downside of the "cache plan in SQL from becoming bloated"? Does it not evict old cache entries?

mak0t0san:
I'm not a SQL expert either but our eagle-eyed DBA caught this while analyzing performance issues.  SQL compiles a query and saves the plan that is generated in cache so that the same query can be re-ran without generating the plan again which optimizes subsequent queries.  Since the query is different (due to the @payload parameter changing), it looks like a different query and then it gets recompile and another cache entry is created.

There can be 10s of thousands of these, each around 400k in size, which takes up more memory, causes cache misses, and takes longer for the cache lookup.

I've reached out to our DBA who is currently on vacation and requested for his feedback on this issue. Hopefully he can provide a better answer than mine.

I do know that it was a big enough problem that stood out for him to notice, and I dug into what was causing the issue.

RemingtonBlake:
There is indeed a process for aging out plans in cache based on relevance, use etc. The compiling and caching of plans over and over again for the (nearly) same query is not ideal as it’s an additional expenditure of CPU and memory. As the data type lengths are different each time on these, SQL views each individual request as a unique and novel query and treats them as such. Usually you see cache bloat from queries using literal predicates instead of parameterized predicates. Enabling Forced Parameterization at the database level can help mitigate that problem. I tried that here to no avail – which makes sense as data type lengths are not literal predicates. 

It’s been awhile since I’ve looked at this, but I am not currently seeing bloat from SignalR. I’m not sure why – possibly due to our upgrade to SQL ‘19. Cache bloat is a persistent problem in our system, but it has been hard to make headway on and to be honest I’m not sure how much damage it’s causing. A lot of the bloat is coming from ORM queries. Those are not so easily extracted from the app code, put into SQL, and parameterized etc. 

One concern I would have in making this a varchar(max) is that memory grants could be negatively affected – SQL would think it’s churning through a lot more data than it actually is every time for one of these and over grant memory.

src/Microsoft.AspNet.SignalR.SqlServer/SqlSender.cs
```diff
@@ -43,14 +42,17 @@ public Task Send(IList<Message> messages)
                 return TaskAsyncHelper.Empty;
             }
 
-            var parameter = _dbProviderFactory.CreateParameter();
-            parameter.ParameterName = "Payload";
-            parameter.DbType = DbType.Binary;
-            parameter.Value = SqlPayload.ToBytes(messages);
+            using (var stream = new MemoryStream(SqlPayload.ToBytes(messages)))
```
roji, line 45: IIRC rather than wrapping `byte[]` with MemoryStream, you can simply set `Size = int.MaxValue` to get the same effect; the parameter should get sent as a `varbinary(max)`.

roji (commented):
This looks correct to me - binary parameters with varying sizes would cause multiple cached plan entries, which means SQL Server is recompiling every time and possibly evicting plans for other important queries.

See suggestion for improvement for below - a wrapping MemoryStream shouldn't be necessary here.

roji:
One note: I'd indeed give this a thorough test/benchmark, since SqlClient has been known to have issues around binary/large data (see https://github.com/dotnet/SqlClient/issues/593 for the read side), and this changes the transport mechanism within SqlClient (I think to "partially length prefixed", PLP).

@cheenamalhotra or @Wraith2 may have more insight here.

Wraith2:
If you have the option to use a Stream rather than materializing the string or binary into an array It's worth considering doing so. The current internal details of how multi-packet payloads are reconstructed inside SqlClient mean that it will have O(n!) time not the O(n) you would expect, using a stream avoids the intermediate array and lets you stream from source to sink as data arrives.

roji:
@Wraith2 as far as I can tell, the source data is already a byte[] in memory rather than something streamable (e.g. a file); this PR proposes to wrap the existing byte[] with a MemoryStream just as a way to make SqlClient send it as a `varbinary(max)` (instead of `varbinary(100)` or whatever).

So the question is more this - given a set of existing byte[] instances in memory, sending them as a different `varbinary(x)` each time causes query cache pollution on the SQL Server side. Do you think there's any issue on the SqlClient side with sending them as `varbinary(max)` (and therefore as PLP)?

Wraith2:
Sending them as type varbinary(max) should be fine.

halter73:
Given that people who know SQL Server better than I do aren't telling us this is a critical change we must do and are instead suggesting that it should be fine and we should benchmark, I don't think this meets our bug bar.

> IMPORTANT: There will be no new features added to this product. We are in maintenance mode and only fixing critical issues. See our blog post for details. This repository hosts code and project management for ASP.NET SignalR, for use in .NET Framework applications using System.Web or Katana. If you are looking for information on ASP.NET Core SignalR, see the aspnet/AspNetCore@master/src/SignalR repository.

https://github.com/SignalR/SignalR#aspnet-signalr

Issue #4416, opened 2019-10-09
Lock contention in TransportHeartBeat.GetConnections()


paveliak:
We had an incident and while analyzing Perfview trace from it I saw lock contention on the ConcurrentDictionary.GetValues() method that leads to [TransportHeartBeat.GetConnections](
https://github.com/SignalR/SignalR/blob/61c1a688f6a4e714e43c5b82b10f6d8b7c612dd6/src/Microsoft.AspNet.SignalR.Core/Transports/TransportHeartBeat.cs#L169)

Changing the code in this way should fix it 
```
return _connections.Select(x => x.Value.Connection).ToList();
```
### Expected  behavior
No lock contention

### Actual behavior
Code acquires all locks of the underlying concurrent dictionary causing lock contention under high load

davidfowl:
@paveliak do you care to send a pull request for this?

paveliak:
@davidfowl I tried but `git push` fails with 403 saying that permission is denied for me

davidfowl:
Yes you can’t push to this repository, you have to fork

paveliak:
Here is the PR https://github.com/SignalR/SignalR/pull/4469

PR #4469, opened 2020-05-06
Fixed lock contention when querying connections

paveliak:
Something that we noticed while analyzing Perfview trace taken during the incident with our service. Perfview showed that there was a lock contention on the `ConcurrentDictionary.Values` property that pointed to `TransportHeartBeat.GetConnections()` method.

`ConcurrentDictionary.Values` would create a snapshot of the dictionary which requires taking global lock of all dictionary buckets. We should not need snapshot behavior here and can use "lock-free" enumeration of the dictionary members.

This is a fix for https://github.com/SignalR/SignalR/issues/4416


halter73:
Thanks @paveliak!

PR #4475, opened 2020-05-22
readme updates

bradygaster:
per the doc issue reported [here](https://github.com/dotnet/aspnetcore/issues/21618), this update makes it more up-front which client customers should use, when their only knowledge of SignalR Is from perusing NPM packages, or who find their way here via OG SignalR docs.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/README.md
+++ b/src/Microsoft.AspNet.SignalR.JS/README.md
@@ -1,6 +1,10 @@
 # ASP.NET SignalR JavaScript Client
 
-This is the client for ASP.NET SignalR. Use this package if you are connecting to an application written in ASP.NET (using System.Web, or OWIN self-host). This package is **not compatible** with ASP.NET **Core** SignalR. If you are connecting to an ASP.NET Core application, use the `@microsoft/signalr` package.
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.SignalR.JS/README.md
```diff
@@ -1,6 +1,10 @@
 # ASP.NET SignalR JavaScript Client
 
-This is the client for ASP.NET SignalR. Use this package if you are connecting to an application written in ASP.NET (using System.Web, or OWIN self-host). This package is **not compatible** with ASP.NET **Core** SignalR. If you are connecting to an ASP.NET Core application, use the `@microsoft/signalr` package.
+This is the client for ASP.NET SignalR. Use this package **only** if you are connecting to an application written in ASP.NET (using System.Web, or OWIN self-host).
+
+> Note: This package is **not compatible** with ASP.NET **Core** SignalR and is not the recommended client for building serverless applications with Azure SignalR Service.
```
halter73, line 5: Will this show up like "Note: The file extension .md indicates a Markdown file." does on https://docs.npmjs.com/about-package-readme-files? If so, great!

halter73: It might be best to put this at the very top, even above the "# ASP.NET SignalR JavaScript Client	# ASP.NET SignalR JavaScript Client" header.

Issue #4476, opened 2020-05-27
ASP.NET SignalR JavaScript Library version number in hubs.js


Staminah:
Hi, 

Would it be possible to update version number indicated in the line `ASP.NET SignalR JavaScript Library v2.3.0-rtm` to `v2.4.1` in the following file : [https://github.com/SignalR/SignalR/blob/2.4.1/src/Microsoft.AspNet.SignalR.Core/Scripts/hubs.js](url)

Because of this, some tools that analyzes the code detects this and indicates that the version used of SignalR is not up to date.

### Expected  behavior
The version number specified in the header comment is matching with the current branch tag (2.4.1).

### Actual behavior
The version number specified in the header comment is not matching with the current branch tag.

### Steps to reproduce
--

Thanks in advance.

Best regards

BrennanConroy:
@halter73

yguerdat:
Hello @BrennanConroy 

Do you have any plan for the next release ? We need to inform our customer because the wrong version's number generated create an issue in the security report of our partner. 

Thanks for your feedback and have a nice day

BrennanConroy:
I don't have a timeline for the next release sorry. But this will be fixed in 2.4.2

Fixed by https://github.com/SignalR/SignalR/pull/4477

PR #4477, opened 2020-05-29
Update hubs.js versions automatically

halter73:
Addresses #4476


samples/Microsoft.AspNet.SignalR.Samples/Scripts/hubs.js
```diff
@@ -1,720 +0,0 @@
-/*!
```
BrennanConroy, line 1: Does this break the sample?

halter73: No. The file was already auto generated at build time. I deleted it and added it to .gitignore because I didn't want to have to update this file every time the version changes.

PR #4572, opened 2021-07-15
Redirect users to the latest version docs

spidyhackx:


BrennanConroy:
ASP.NET SignalR and ASP.NET Core SignalR are two different products. So the docs should not point to the other one.

PR #4594, opened 2021-10-07
Migrate to 1ES hosted pools

lpatalas:
We are migrating all repositories to 1ES hosted pools. The functionality stays the same but we have to update the yamls.

Tracking issue: https://github.com/dotnet/core-eng/issues/14276

/cc: @jonfortescue


BrennanConroy:
Public worked, I'm going to assume internal will too, worst case we can fix it up 😄

BrennanConroy (approved):

PR #4620, opened 2022-02-17
Add SBom generation

BrennanConroy:
Part of https://github.com/dotnet/arcade/issues/8477

### DIFF
--- a/.vsts-pipelines/templates/signalr-build.yml
+++ b/.vsts-pipelines/templates/signalr-build.yml
@@ -67,4 +67,7 @@ jobs:
       buildConfiguration: $(BuildConfiguration)
       buildPlatform: $(AgentOsName)
 
+- ${{ if and(eq(parameters.runAsPublic, 'false'), ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+  - template: /eng/common/templates/steps/generate-sbom.yml
+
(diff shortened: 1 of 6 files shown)


.vsts-pipelines/templates/signalr-build.yml
```diff
@@ -67,4 +67,7 @@ jobs:
       buildConfiguration: $(BuildConfiguration)
       buildPlatform: $(AgentOsName)
 
+- ${{ if and(eq(parameters.runAsPublic, 'false'), ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
```
halter73, line 70: Why skip in these conditions? Wouldn't it make more sense to just not retain the build in that case? Why aren't we using retain-build.yml?

halter73 (commented):
Approving to unblock.

.vsts-pipelines/templates/signalr-build.yml
BrennanConroy: Oh, didn't see this link: https://github.com/dotnet/arcade/blob/main/Documentation/SBOMGenerationGuidance.md#retention-rules-for-release-build-sboms
I'll look into that before merging

Copied the condition from arcade: https://github.com/dotnet/arcade/blob/78eaf78761027d225030be2b28aaf4e8bf392929/eng/common/templates/job/job.yml#L256-L257

Issue #4587, opened 2021-09-03
Client unable to receive message with Jquery v3.6.0


eeldivad:
### Expected  behavior
I have Chrome opened on my Windows 10 machine and Chrome opened on my Windows server 2019 machine.  When I type a message and click send, it sends to the hub service and the hub service simply calls
Clients.All.addMessage(message);

So this should tell all clients to display the message.

### Actual behavior
It only sends the message to Windows 10 Chrome machines, but not any of my Windows Server 2019 with Chrome.  I tried multiple combinations and the message never arrives on Server 2019.  

### Steps to reproduce
`
Create a simple hub method
        public void Send(string message) {
            Clients.All.addMessage(message);
        }

Clientside code:
            testHub.on("addMessage", function (message) {
                $('#logs').append(message);
            });
`
- Chrome Version 93.0.4577.63 (Official Build) (64-bit)
- jquery.signalR.js version 2.4.2
- jquery 3.6.0
- Open Chrome browser on Windows 10 and another one on Windows Server 2019
- You will notice the message reaches the Windows 10 machine but not Server 2019

If I downgrade jquery to v2.1.3 then it works as expected.  Please fix issue related to latest jquery 3.6.0

BrennanConroy:
We haven't seen other reports of this and given it seems to be machine specific we will close for now. If there are more reports of this sort of problem or if there are more in-depth details we can reopen.

PR #4625, opened 2022-05-02
Upgrade JQuery to fix CG alerts

wtgodbe:
Update JQuery to 1.9.0


BrennanConroy:
Yeah, we can't do this... @halter73

wtgodbe:
> Yeah, we can't do this...

Updating to 1.9.0 would fix some, but not all alerts. We'd have to get exemptions for the remaining ones (which are all `Medium` severity), which is possible if we have a good justification

wtgodbe:
Surprisingly enough, the build passed (The only project referencing the `jquery` package directly is `Microsoft.AspNet.SignalR.JS`, which built fine)

halter73:
Forcing an upgrade to multiple major versions of jQuery is *very* risky. jQuery has undergone some big changes since 1.6.4. We do have automated headless-browser-hosted tests and I believe upgrading event to 3.6.0 mostly works, but we've had reports of SignalR breaking with 3.x versions of jQuery on specific platform combinations we do not have automated tests for https://github.com/SignalR/SignalR/issues/4587.

We'd need to invest resources to manually test more client/server combinations in more scenarios before I'd feel comfortable making this kind of change. It's possible updating jQuery is only affecting certain transports on certain browsers, but I'd want to know before forcing an update.

We also don't have any evidence that the way SignalR uses older versions of jQuery is itself insecure. Developers are encouraged to upgrade jQuery to the latest version if that works for their app and use case, but we have to consider that jQuery has a much bigger API surface than this SignalR JS client which hasn't had a major API change in a long time.

If there were a security patch we needed to make to this SignalR JS client, forcing app authors to deal with multiple major versions worth of breaking changes in the jQuery. This would leave a substantial portion of the consumers of this library out in the cold. We have to consider that developers using this library still haven't updated to ASP.NET Core, so we're talking about a lot of legacy apps that haven't been rewritten due to breaking API changes.

BrennanConroy:
Why 1.9 instead of newer 1.x versions?

Also https://github.com/SignalR/SignalR/blob/main/src/Microsoft.AspNet.SignalR.JS/Microsoft.AspNet.SignalR.JS.nuspec#L19 needs updating.

wtgodbe:
> Why 1.9 instead of newer 1.x versions?

1.9.0 is the minimum version to satisfy CG. 1.12.4 was resolved in the package-lock, we could go with that instead.

Issue #4451, opened 2020-03-11
Stop method doesn't respect waitForPageLoad passed by user


michax:
Stop method is checking waitForPageLoad config property, however this config property from connection is always deleted before we check for it's existence and it's properties like waitForPageLoad.

Here's where config is deleted:
https://github.com/SignalR/SignalR/blob/9ab6410822b6a877590ded2e5185b3ad3ade4932/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js#L980

And few lines below we check for config existence and state of waitForPageLoad. Config in this place will never exists:
https://github.com/SignalR/SignalR/blob/9ab6410822b6a877590ded2e5185b3ad3ade4932/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js#L985

Summary of the problem:
```
// Config being deleted
delete connection._.config;
delete connection._.deferredStartHandler;

// Below check for waitForPageLoad state will never be respected because connection._.config will always be undefined here.
if (!_pageLoaded && (!connection._.config || connection._.config.waitForPageLoad === true)) {...}
```

PR #4646, opened 2022-09-27
Update jquery.signalR.core.js

mkorganashvili:
"config" is cleared before it is checked in "if" condition and skips disconnection if page is still loading

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
@@ -979,7 +979,6 @@
             }
 
             // Always clean up private non-timeout based state.
-            delete connection._.config;
             delete connection._.deferredStartHandler;
 
             // This needs to be checked despite the connection state because a connection start can be deferred until page load.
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Issue https://github.com/SignalR/SignalR/issues/4451

src/Microsoft.AspNet.SignalR.JS/jquery.signalR.core.js
```diff
@@ -979,7 +979,6 @@
             }
 
             // Always clean up private non-timeout based state.
-            delete connection._.config;
```
BrennanConroy, line 982: Could we instead move the `(!connection._.config || connection._.config.waitForPageLoad === true)` code and assign it to a local variable before deleting config? That way we don't have to make sure we delete config in all conditions below.

mkorganashvili: Well, good idea

BrennanConroy (approved):

BrennanConroy:
Thanks for the change!

Issue #4386, opened 2019-06-25
off() called with a non-null callback will never remove the registration from the list or delete the callbackMap


Mythagoz:
### Expected  behavior
Calling .off() with a non-null callback should that callback registration from the list and delete the callbackMap is that registration was the last remaining registration.

### Actual behavior
In `.off()`, the callback index to remove, `callbackIndex`, is determined in the following loop that will leave `i` at `callbackSpace.length`:
```javascript
for (var i = 0; i < callbackSpace.length; i++) {
    if (callbackSpace[i].guid === callbackIdentity._signalRGuid) {
        callbackIndex = i;
        callbackRegistration = callbackSpace[i];
    }
}
```

When it comes time to splice the array, `i` is used instead of callbackIndex.  As `i` is `callbackSpace.length`, it has no effect on the array:
```javascript
// Remove the registration from the list
callbackSpace.splice(i, 1);
```

`callbackMap` will never be deleted as the code checks for `callbackSpace.length` to be changed to 0:
```javascript
// Check if there are any registrations left, if not we need to destroy it.
if (callbackSpace.length === 0) {
    delete callbackMap[eventName];
} 
```
This can result in memory leaks in code that depends on `callbackMap` being cleaned up.

### Steps to reproduce
Call .off() with a non-null callback

nikitaky16:
Сonfirm that, I encountered the same problem.

BrennanConroy:
Fixed via https://github.com/SignalR/SignalR/pull/4659

PR #4659, opened 2022-12-27
fix hub.off function has a wrong "unsubscription"

Mariachi1231:
Fix for the bug introduced in #4249 ,
also mentioned in #4386.

### DIFF
--- a/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
+++ b/src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
@@ -175,7 +175,7 @@
             // Verify that there is an event space to unbind
             if (callbackSpace) {
 
-                if (callback) {
+                if (callbackIdentity) {
                     // Find the callback registration
                     var callbackRegistration;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.SignalR.JS/jquery.signalR.hubs.js
```diff
@@ -194,14 +196,15 @@
...
+                        callbackSpace.splice(callbackIndex, 1);
 
                         // Check if there are any registrations left, if not we need to destroy it.
                         if (callbackSpace.length === 0) {
                             delete callbackMap[eventName];
                         }
                     }
-                } else if (!callback) { // Check if we're removing the whole event and we didn't error because of an invalid callback
+                }
+                else if (!callbackIdentity) { // Check if we're removing the whole event and we didn't error because of an invalid callback
```
halter73, line 207:
```suggestion
                } else if (!callbackIdentity) { // Check if we're removing the whole event and we didn't error because of an invalid callback
```

halter73 (approved):
Thanks!

PR #4662, opened 2023-01-23
Swallow unhandled rejection explicitly in ajax transport when using fetch

benjamingr:
At Excel Online we've noticed errors showing up in instrumentation when fetch fails to fetch when stopping a connection (e.g. on page unload).

This PR swallows that error explicitly instead of implicitly. Also good: actually handling the error would also meet our needs.

Cheers, love and thanks


benjamingr:
piing

eliassultan778 (approved):

PR #4713, opened 2024-07-12
dependency updates

danmoseley:
Just ran the tool to update all (nuget) package versions as some have CVE's.

This is a useless PR because 
1) more work would be needed to make it build, eg., move net45 to net462
2) we surely don't want all these updates, some of which are big leaps, but perhaps only those that have CVEs. (It remains the reponsibility of the app using SignalR to have updated dependencies, but it's nice for SignalR to reference them.)

CG dashboard showing the interesting subset
https://msit.powerbi.com/groups/me/reports/7c80142e-fa00-47d3-a63a-a3b6b338e15c/ReportSection?experience=power-bi

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,45 +1,45 @@
 <Project>
   <PropertyGroup Label="Package Versions: Auto">
-    <CastleCorePackageVersion>4.0.0</CastleCorePackageVersion>
+    <CastleCorePackageVersion>5.1.1</CastleCorePackageVersion>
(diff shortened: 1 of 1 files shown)


danmoseley:
Confirmed with Levi that while we should keep the repo up to date with any dependencies that have fixed vulnerabilities, we do not ship a release simply to bring those, as it's the app that ultimately chooses versions anyway. IOW -- this is not urgent, but some of these are required to be updated at some point for whenever we release for some other reason.
cc @BrennanConroy

PR #4764, opened 2026-05-27
Update unload event to pagehide in SignalR (Chrome deprecating unload…

gjamio:
Enhance unload event handling by checking if the page is persisted before stopping the connection. - unload event in documents are deprecated [https://developer.chrome.com/docs/web-platform/deprecating-unload](https://developer.chrome.com/docs/web-platform/deprecating-unload)


maxschaf:
we see this issue in prod now, any plans to fix it?

michaels-intact:
Hey @halter73, seems like you are the maintainer of this, since you approved the last PRs.
Could you maybe take a look?

